Issue #10, opened 2015-10-09
Provide a robust fix for unknown enums


nozzlegear:
Shopify's documentation comes nowhere close to documenting all available enum values, and this has really thrown a wrench in things. ShopifySharp tries to provide enums where applicable, but I've personally run into exceptions where Shopify returns an enum that isn't documented (therefore isn't in ShopifySharp) and fails to deserialize.

I feel that enums are still preferable to using plain strings, so we need a more robust solution that won't throw exceptions when trying to deserialize unknown enums.

Instead, the current plan is to make all of ShopifySharp's enums implement an "Unknown" value, and make them all use a custom converter that will deserialize unknown enum strings into the "Unknown" value. The ShopifySharp.Enums.ShopifyProcessingMethod is the first to do this.

nozzlegear:
Rather than making each enum implement an "Unknown" value, it's probably better to just make them all nullable. Checking if an enum is null takes the same effort as checking if it's "unknown". To avoid breaking changes, we'll have to save fixing this for existing enums until a v 2.0 release. 

Enums created from now on, though, will be nullable.

nozzlegear:
TODO

- [] Convert the `ShopifyFulfillment.Status` property type to a nullable enum.

mdossantos:
Hi team, just curious if this fix will be available soon? We are currently having issues with it and hoping that it will be resolved soon before we jump to another package or workaround. thanks

nozzlegear:
Hey @mdossantos, I'm looking at starting this soon but can't make any promises on exactly when it would be ready. Could you list which enums are specifically giving you trouble? I'd be happy to push a small update to fix it.

mdossantos:
Hi Joshua,

Thanks for the quick reply. The enum we are having trouble with is
ShopifySharp.Enums.ShopifyProductFulfillmentService.
We get error converting the type "xxxxFulfillmentServiceNamexxxx" to
ShopifySharp.Enums.ShopifyProductFulfillmentService.
We get the error anytime our products are from a different fulfillment
service from shopify that isn't 'Manual'. This is when grabbing the list of
products from shopify.
Thanks

nozzlegear:
Thanks! I've just opened up an issue (#41) to track this. The fix will be switching `ShopifyProductFulfillmentService` to a string because there's no way to track all of the custom fulfillment services that apps and developers can create. I'll try to get it in this weekend if my schedule cooperates.

mdossantos:
[...]

PR #15, opened 2015-11-11
ShopifyThemeService: Create, retrieve, list, update and delete a store's themes.

nozzlegear:
- Merges the ShopifyAssetService into master, which enables creating, retrieving, listing, updating and deleting a store's themes. 
- Bumps the version to 1.10.0.
- Readme has been updated with usage examples.
- This PR merges the first use of `NullableEnumConverter<T>`, which is the final solution for [handling Shopify's undocumented and sometimes-null-sometimes-not enum values](https://github.com/nozzlegear/ShopifySharp/issues/10). Nullable enums are not backwards-compatible, so we'll wait until version 2.0 before switching all previous enums to this converter.


nozzlegear:
All tests are passing.

PR #18, opened 2015-11-27
Updated the "password_enabled" field for the shop object making it nullable

squallop:
Contrary to Shopify API documentation, the "password_enabled" for on a shop object is a nullable boolean value. Please correct me if I'm wrong :)

### DIFF
--- a/ShopifySharp/Entities/ShopifyShop.cs
+++ b/ShopifySharp/Entities/ShopifyShop.cs
@@ -139,7 +139,7 @@ namespace ShopifySharp
         /// Indicates whether the Storefront password protection is enabled.
         /// </summary>
         [JsonProperty("password_enabled")]
-        public bool PasswordEnabled { get; set; }
+        public bool? PasswordEnabled { get; set; }
 
         /// <summary>
         /// The contact phone number for the shop.


squallop:
Sorry I should have also noted that this causes a JsonSerializationException in Newtonsoft.Json when retrieving a shop using the ShopifySharp.ShopifyShopService / GetAsync() method, which is how I found it.

nozzlegear:
Thanks Shannan!

It's not a stretch to believe that Shopify's documentation is wrong on this. I can't count the number of times that I've had something break because their docs say they'll send X value, but make no mention that the value might also be null, or it might be one of 20 other undocumented values and you just need to figure it out as you go along. 

That might fly in dynamic Ruby land or whatever they're using, but it breaks things for us C# devs.

Anyway, I'll get this merged and bump the version either later tonight or over the weekend. Thanks for contributing!

nozzlegear:
I've merged this PR and updated the NuGet package to 1.11.3 with this change. 

Thanks again!

squallop:
No problems at all, thanks Josh.

Keep up the good work, it's a great project!

PR #19, opened 2015-12-17
Adding since_id for OrderFilterOptions

Yitzchok:
Add support for since_id for orders

https://docs.shopify.com/api/order has since_id


nozzlegear:
Thanks Yitzchok! Not sure how I missed this one. I've merged your PR and will get the NuGet package updated soon.

Yitzchok:
Thanks

Btw I think at one point it was missing from the shopify docs for order.

nozzlegear:
Alright, that makes sense then. I've updated the NuGet package to 1.11.4 with this change, it should be available in the next few minutes. Thanks again!

PR #20, opened 2015-12-19
Shopify collect service

stepankobzey:
Added support for collect mapping products to collections


nozzlegear:
Thanks Stepan! This is looking good. I need to ask for one change to bring the "style" in line with the rest of the library, though, before I can merge this.

Could you change the casing for the properties in **ShopifySharp/Entities/ShopifyCollect.cs** and **ShopifySharp/Services/Collect/ShopifyCollectFilterOptions.cs** to C#'s PascalCasing? Then, to ensure the models still work with the Shopify API, add a `JsonProperty` attribute to each property, using your current casing.

For example, the `ShopifyCollect` model would look like this:

```cs
public class ShopifyCollect: ShopifyObject
{
    /// <summary>
    /// The id of the custom collection containing the product.
    /// </summary>
    [JsonProperty("collection_id")]
    public long CollectionId { get; set; }

    /// <summary>
    /// The unique numeric identifier for the product in the custom collection.
    /// </summary>
    [JsonProperty("product_id")]
    public long ProductId { get; set; }

    ...
}
```

I'll also need some test coverage for each of the `ShopifyCollectService` methods, but I'd be happy to write those myself if you prefer.

stepankobzey:
should be good now. didn't made any tests, its runs heavily in my system now

nozzlegear:
Great, thanks! I'm on holiday vacation right now, so I haven't had a chance to get this merged and updated yet. Might not get a chance until I'm back next week.

nozzlegear:
Looks like I had more time tonight than I thought. I've got your PR merged in and wrote some tests for all of the service methods.

There were a few bugs in the service that I fixed: the `CreateAsync`, `ListAsync` and `CountAsync methods weren't deserializing their results, instead returning collects with empty values. I also changed the name of `ShopifyCollectService.Delete` to `ShopifyCollectService.DeleteAsync`. 

If you're using those methods in production, I'd recommend that you update to the latest NuGet version or manually edit the code.

Anyway, all tests are passing and the NuGet package has been updated to v1.12.0. Thanks for contributing, Stepan!

PR #24, opened 2015-12-26
ShopifyTransactionService: Create, get, list and count an order's transactions.

nozzlegear:
- Adds support for the `ShopifyTransactionService`.


nozzlegear:
All tests passing.

PR #25, opened 2016-02-03
Added ShopifyPageService to control pages

Panksy:
This allows the create, update and deletion of pages. It also allows the creation and updating of page metafields.

Unit tests have also been added.


nozzlegear:
Thanks for the PR, Cathy! This is some good work here and I don't see any issues with it. I'll make sure it gets merged in and update the Nuget package this weekend.

Panksy:
Great, if you are happy with that I'll add in the Service for the blog articles at some point too. I like what you've started here.

nozzlegear:
That would be awesome! I've got this PR merged into master and your `ShopifyPageService` should now be available on Nuget, version 1.15.0.

Issue #21, opened 2015-12-25
Clean up the filter classes for .Count and .List commands.


nozzlegear:
The list/count filter classes have a bunch of duplication going on. List filter classes should instead be replaced by a generic `ShopifyListFilter` class that contains the following props:

```cs
public long SinceId

public int Page

public string Fields
```

It will inherit from a `ShopifyCountFilter` class, which contains the following props:

```cs
public DateTime CreatedAtMin

public DateTime CreatedAtMax

public DateTime UpdatedAtMin

public DateTime UpdatedAtMax
```

Almost every count and list command uses the properties in these two classes. The exceptions are those that either can't be filtered, or those that use all of these properties **plus** a couple of custom properties that are specific to their objects. Instead of doing something sensible, we've somehow (entirely my fault!) ended up with a bunch of different classes that repeat these common properties.

This would be a breaking change, though, so I'm marking it for Version 2.

PR #27, opened 2016-02-03
adding OrderFilterOptions

mooglegiant:
adding in the options supported by the shopify API


nozzlegear:
Thanks for the PR @mooglegiant! I had thought these filter options were already in the lib, but it looks like the OrderService is suffering from the problem described in issue #21. I'll get your changes merged in this weekend.

mooglegiant:
I don't really mind if the different filter objects are redundant.  I just started a new project that needs Shopify integration and just spent the last two days trying to understand their API.  Your wrapper makes it so much easier to use. They already have a process for storing the last OrderID processed. so I can use the GetSince method.  I just noticed it was missing and figured I would add it.

nozzlegear:
I've got your fixes merged into the master branch, and they should now be available on Nuget with version 1.15.0. Thanks again!

PR #35, opened 2016-02-11
Two New Services (only data pulling from shopify)

ishahrier:
I have added two new services
1. Custom Collection Services
2. Smart Collection Service
(only the getter methods)

I am using these two services with no issues in my own projects. Once i will have more time i will update these services by providing the create and delete methods a well.

thanks


ishahrier:
class name is not right..i will rename the classes..they have "My" prefix in there

nozzlegear:
No problem @ishahrier! You should be able to push commits to the same branch and they would be added to any open pull request.

PR #36, opened 2016-02-18
Unpublish a product logic

mooglegiant:
To unpublish a product from the API, you must be able to set the published_at date to null, which requires json serialize attributes to change.


ishahrier:
Awesome!!

nozzlegear:
Thanks for the PR! I've got this merged and it should be available on Nuget as v1.15.1 within the next hour. I've also added `PublishAsync` and `UnpublishAsync` methods to the `ShopifyProductService`, because there was one case where just setting `published_at` to null and updating the product would throw an error.

PR #39, opened 2016-03-08
Added support for Custom Collections Service

damazoarriaga:
Added support for Custom Collections Service and updated Custom Collections to include a List of Collects


nozzlegear:
Thanks Angel, this looks good! I'll get it merged in as soon as I get the chance, though it may not be until later this week or this weekend.

nozzlegear:
Thanks again, Angel. I've added some tests and the `ShopifyCustomCollectionService` should be available on Nuget with v1.17.0 within the next hour.

I did have to make one change: the `ShopifyCustomCollection.Image` property was previously a string, but it's now its own object to match up with the Shopify API.

Issue #41, opened 2016-03-24
ShopifyProductFulfillmentService should be a string.


nozzlegear:
As reported in #10, [`ShopifyProductFulfillmentService`](https://github.com/nozzlegear/ShopifySharp/blob/eebede3342e7e55e10b7c914b091c84bc12e70c8/ShopifySharp/Entities/ShopifyProductVariant.cs#L50) should not be an enum. The Shopify API lets apps and developers create their own fulfillment services which could never be mapped to a hard-coded enum. While we could switch this over to a nullable enum, that's not very useful for developers trying to specify their own fulfillment service too. 

To fix, this enum should be converted to a string.

Issue #14, opened 2015-11-05
Update ShopifyAuthorizationService.IsValidRequest


nozzlegear:
Shopify has recently deprecated the MD5 + `signature` request validation method, replacing it instead with SHA256 + `hmac`. While MD5 + `signature` still works, it's only a matter of time before Shopify removes it completely.

nozzlegear:
**Urgent**: Current implementation will break on [June 1st, 2016](https://docs.shopify.com/api/guides/authentication/oauth#making-authenticated-requests).

Issue #9, opened 2015-10-09
Return Shopify API errors as KeyValuePairs instead of a JSON string


nozzlegear:
Right now, ShopifySharp returns any errors returned by the Shopify API as a JSON string. The errors are formatted like this:

```
{
    "errors": {
        "addresses.country": [
            "is not a valid country"
        ]
    }
}
```

It would be better if they were returned as a list of KeyValuePairs, which would allow the developer to list them out and sort them by their type.

nozzlegear:
This change is not backwards compatible, and will need to be implemented in version 2.0.

nozzlegear:
**Urgent**: Current implementation will break on [June 1st, 2016](https://docs.shopify.com/api/guides/authentication/oauth#making-authenticated-requests).

PR #43, opened 2016-04-04
Version 2.0.0

nozzlegear:
This PR will bump the library to version 2.0.0 and contains several breaking changes.

**To-do**

- [x] ShopifyProductFulfillmentService should be a string. #41.
- [x] All `*FilterOptions` and `*ListOptions` should inherit from the base [`ShopifyListFilter`](https://github.com/nozzlegear/ShopifySharp/blob/3ddc3f4/ShopifySharp/Services/ShopifyListFilter.cs) and [`ShopifyCountFilter`](https://github.com/nozzlegear/ShopifySharp/blob/3ddc3f4/ShopifySharp/Services/ShopifyCountFilter.cs) classes. #21.
- [x] Update `ShopifyAuthorizationService.IsValidRequest`. #14.
    - [x] Does `IsAuthenticProxyRequest` use the same method? If so, can be completely removed. Edit: **no**. `IsAuthenticProxyRequest` still uses a `signature`, doesn't send an `hmac`.
- [x] All enums should be nullable and implement `NullableEnumConverter<T>`. #10.
- [x] Return Shopify errors as a KVP list instead of a JSON string. If error cannot be deserialized, return the JSON string as the first KVP (e.g. `{ "genericError" : "json string error" }`). #9.
- [x] [Uninstall an app via API](https://docs.shopify.com/api/guides/authentication/uninstalling-applications) (`ShopifyShopService.UninstallAppAsync`).
- [x] ~~List current app permissions.~~ Not possible. `https://someshop.myshopify.com/admin/api_permissions/current.json` returns a 403 unauthorized response.
- [x] Change `ShopifyOrder.NoteAttributes` to a list of `ShopifyNoteAttribute` and allow any kind of object value. #44.


nozzlegear:
Some release notes as I go along:

- `ShopifyException.Error.Errors` is now a `Dictionary<string, IEnumerable<string>>` on the ShopifyException itself. To maintain some back compat, `ShopifyException.JsonError` is the raw JSON-serialized error returned by Shopify. It's functionally identical to ex.Error.Errors.

nozzlegear:
v2.0.0 should be ready to merge into master. Just need to manually test the `ShopifyShopService.UninstallAppAsync` later this evening or tomorrow morning.

PR #48, opened 2016-04-16
adding a null to zero convert for gram field in ShopifyProductVariant

shajumohamed:
adding a null to zero converter to gram field in ShopifyProductVariant class  since the service response had the field as null causing an exception


nozzlegear:
Thanks! Looks pretty simple and shouldn't be a problem to merge in. Just thinking here, would it be better to make the grams field a nullable int instead? I know there's some precedence already with the `FalseToNull` converter, but it might be easier for everyone to just make these nullable.

shajumohamed:
I did that at first. But felt like this implementation will prevent the need of any code change required  on the users' side to handle the nullable int. 

But this does convert a null value to zero. I am not sure whether that will cause any issues.  

Your call :)

nozzlegear:
Alright, while I'd like to change these fields to nullable int/boolean in the next major release, I don't see any problem with using special converters for now. 

Thanks for contributing! Your change should be available on Nuget at v2.0.2.

PR #55, opened 2016-06-06
Adding LineItem Properties

Yitzchok:


nozzlegear:
Thanks for this @Yitzchok! It should be available on Nuget with ShopifySharp v2.1.2.

PR #56, opened 2016-06-08
TotalWeight should be a long instead of an int.

Yitzchok:


nozzlegear:
Thanks! This is now available in v.2.1.3 on Nuget.

Yitzchok:
Great thanks

PR #57, opened 2016-06-20
Add missing authorization scopes read_analytics, read_users and write…

Yitzchok:
…_users


nozzlegear:
Thanks! Normally I get an email for new auth scopes but I couldn't find one for these, I must have missed it. Should be up on Nuget with v2.1.4.

PR #65, opened 2016-09-08
3.0 remove enums and replace them with strings.

### LINKED ISSUE #64: ShopifyOrderStatus Enums need updating
ShopifyOrderStatus is missing some new enums, so I am unable to download unfulfilled orders. I can currently only receive all open orders for a given store

nozzlegear:
This 3.0 release is a major SemVer release, containing a breaking change by removing almost all enums and changing their usage across various classes to a string property. More info on this decision is contained in issue #64.

This will close #64 when merged.


nozzlegear:
Todo:

- [ ] Document the current expected values for each string that replaced the enum.
- [ ] Set up tests for the ShopifyFulfillmentService and ShopifyOrderService for bringing in objects with the new string values.

ShopifySharp/Filters/ShopifyOrderFilter.cs
```diff
@@ -10,21 +10,21 @@ namespace ShopifySharp.Filters
     public class ShopifyOrderFilter : ShopifyListFilter
     {
         /// <summary>
-        /// The status of orders to retrieve. Default is <see cref="ShopifyOrderStatus.Any"/>.
+        /// The status of orders to retrieve. Known values are "open", "closed", "cancelled" and "any" (default).
         /// </summary>
         [JsonProperty("status")]
-        public ShopifyOrderStatus? Status { get; set; } = ShopifyOrderStatus.Any;
+        public string Status { get; set; } = "any";
```
ZubyH: Maybe not default this to "any", as if the developer forgets to populate it, pulling any can be dangerous as it would pull all orders. with some of the shops I've been testing with I've found not all of them Archive their orders because they like visibility of them - defaulting this could cause hundreds of thousands of orders to potentially be pulled

nozzlegear: Good point, thanks!

nozzlegear:
All tests are passing.

MalikDrako:
It would be useful to have string constants in addition to the documentation. You may also be able to reduce breaking existing code if you specify the constants like the enums were, ex:
```
public class ShopifyFulfillmentStatus {
    /// <summary>
    ///     Every line item in the order has been fulfilled.
    /// </summary>
    public const string Fulfilled = "fulfilled",

    /// <summary>
    ///     None of the line items in the order have been fulfilled.
    /// </summary>
    public const string None = null,

    /// <summary>
    ///     At least one line item in the order has been fulfilled.
    /// </summary>
    public const string Partial = "partial",
}
```

ZubyH:
Isn't the point of removing the enums to your ass control back to the developer? Even if string constants were in place, we'd still not be able to react quickly to a shopify interface change

nozzlegear:
That's a pretty elegant solution, @MalikDrako. I'm disappointed that I hadn't considered it before releasing 3.0! 

@ZubyH, this would let us continue to use strings instead of enums as we do now in 3.0, but we'd also be able to use "hardcoded" strings so we don't have to remember the exact values that need to be sent. Then if Shopify were to change a value you could still quickly update your app to use the new strings instead of waiting for an update from ShopifySharp.

It's essentially having a class of static strings that we can use to remember the currently known values.

```cs
string status = ShopifyFulfillmentStatus.Fulfilled; // This is now a static string instead of an enum
var filter = new ShopifyFulfillmentFilter()
{
  status = status
};
```

Then if Shopify were to change the values of Fulfillment statuses, you could quickly update your app without waiting for an update from ShopifySharp because the filter still accepts strings:

```cs
string status = "updated_value";
var filter = new ShopifyFulfillmentFilter()
{
  status = status
}
```

PR #66, opened 2016-09-09
Event Support

mrjono1:
Hi @nozzlegear 
This is my first pull request so lets hope I got it all correct.
I have added support for shopify events, I was able to implement almost everything mentioned in the spec here https://help.shopify.com/api/reference/event
Accept the type of events like "Get all the events from a particular product" it just kept returning nothing

I think I need Visual Studio Enterprise to run the test scripts? I was unable to run the test scripts, but I did try to create some anyway.

Jono


nozzlegear:
Thanks @mrjono1! This looks pretty good. I've got a few changes before I can merge it in, just some stuff that I haven't yet had a chance to get written up in the readme.

1. Remove the `ShopifyEventSubjectType` and `ShopifyEventVerb` enums and replace them with string properties on your classes. Enums are super convenient, but unfortunately we've started running into problems with them. Shopify will randomly make unannounced changes to them which can break apps (as seen in #64). If their API were versioned we could stick with enums, but for now I'm going to be moving the lib toward using strings to prevent breaking everybody's apps.
2. Instead of using the [JsonIgnore] property on your `ShopifyEventListFilter` and then handling those properties specifically in your service's private method (which looks pretty sharp!), you should be able to override the `Parameterizable` class' `ToSingleParameter` method. You can see an example of that in the [`ShopifyListFilter`](https://github.com/nozzlegear/ShopifySharp/blob/master/ShopifySharp/Filters/ShopifyListFilter.cs#L46).
3. Instead of a custom `ShopifyEventCountFilter`, try to use the generic `ShopifyCountFilter` instead. I know it has 3 properties that don't apply to the Event API, but I'm planning to break these down even further in a future release. For now we try to keep filters as generic as possible unless they have unique properties, to prevent having 50 different types of filters with the same properties shared between them.

Sorry for the list of changes, it's my fault for not documenting this stuff yet. Let me know if something doesn't make sense or you think there's a better way to do it. I'm happy to hear suggestions!

mrjono1:
I'll make the updates.

In regards to Enums, there might be a way to use both methods, the string property and an Enum if it resolves. This could be a project for another day. I will do the string change for now.

mrjono1:
ready for review

nozzlegear:
Awesome work! I forgot to mention, you should be able to run test suites in any version of Visual Studio. ShopifySharp uses Machine.Specifications for tests, though, which means you also need to install the [MSpec test adapter](https://visualstudiogallery.msdn.microsoft.com/4abcb54b-53b5-4c44-877f-0397556c5c44) for VS. I'll run the tests now that this is merged and fix any problems if necessary.

nozzlegear:
I've made a few changes for the tests and published a release. Your ShopifyEventService should now be available on NuGet with ShopifySharp 2.4.0. Thanks!

PR #86, opened 2016-11-14
Support for cancelling orders

### LINKED ISSUE #85: Support cancelling orders
I haven't been able to find any information on how to cancel orders using this API endpoint:
https://help.shopify.com/api/reference/order#cancel

Am I missing something, or is it not supported?

thj-dk:
Adds support for the [cancel order endpoint](https://help.shopify.com/api/reference/order#cancel).

While the `refund` option parameter works, there's no way of testing it because refund in general is unsupported in ShopifySharp.

While the endpoints returns an order, it appears not to be returning the cancelled order. E.g. the cancelled_at parameter is still null. Hence I'm not returning the order from the `CancelAsync` method to avoid confusion.

Resolves #85


nozzlegear:
Thanks for adding this! It should be available on Nuget with version 3.3.0.

PR #94, opened 2016-12-29
IRequestExecutionPolicy interface

clement911:
Introduced IRequestExecutionPolicy interface to specify centralized logic that should run when executing shopify requests.


clement911:
This PR is intended to be non-breaking.
The default policy will effectively use the same behaviour as before.
The retry policy is opt-in only and can be configured as follows when the program starts:

`RequestEngine.SetExecutionPolicy(new RetryExecutionPolicy());`

Users may want to implement their own policies (e.g. tracing requests/responses...).

clement911:
Ok great.

nozzlegear:
Sorry this took so long! I've got it all merged in and documented. Your changes should be available on Nuget with version 3.5.0.

clement911:
Awesome. Thanks!

PR #103, opened 2017-01-15
Add filename property to ShopifyProductImage

mchandschuh:
The shopify api allows you to set the filename when uploading an image.
This change adds the filename property to the image and only serializes
the value if it is specified by the consumer.

See the second example of posting an image to the shopify api [here](https://help.shopify.com/api/reference/product_image#create).


nozzlegear:
Thanks! I've been swamped all of last week, but I'll be working on ShopifySharp tomorrow and will make sure this gets tested and merged in.

nozzlegear:
Your changes should be available on Nuget with version 3.5.1. Thanks!

PR #116, opened 2017-03-08
Fixed json property typo for ShopifyShop.DisplayPlanName

clement911:
I didn't rename the property itself because it would be a breaking change.
Maybe in v4?


nozzlegear:
Thanks! This is on Nuget with 3.5.5.

PR #125, opened 2017-03-27
Added AuthorizeFullResponse

SentientEntities:
Hello,

I created this method to return a full response object so it can be parsed by the consuming application. Please let me know what you think, I'm still getting to know your code so maybe this is already being done somewhere. Thanks!


SentientEntities:
Also, I wanted to mention my reason for doing this: I didn't see anywhere that the "scope" was being returned. My application creates delegate tokens further downstream, so knowing the scope as we get back a response is important when I generate these additional tokens.

PR #128, opened 2017-04-13
Addition of Discount object and Service

mobbsie:
As discussed, these are my first fork changes including:
1. The object model for the Shopify Discount object as well as the associated Service.
2. An upgrade of the project to .NET 4.5.2 and Newtonsoft.Json from 7.0.1 to 9.0.1.

The latter you don't need to merge if you don't want / need to.  I use this project as a library in another project requiring 4.5.2 and Newtonsoft 9.0.1 - hence the need to upgrade.

### DIFF
--- a/ShopifySharp/Entities/ShopifyDiscount.cs
+++ b/ShopifySharp/Entities/ShopifyDiscount.cs
@@ -1,0 +1,94 @@
+﻿using Newtonsoft.Json;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace ShopifySharp
+{
+    public class ShopifyDiscount : ShopifyObject
+    {
+        /// <summary>
+        /// Specify how the discount's value will be applied to the order. Valid values are:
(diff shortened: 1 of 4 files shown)


nozzlegear:
Great, thanks for doing this! I'll get it merged in and tested over the weekend, but I'll probably leave out the framework upgrade just because it might be a breaking change for some devs.

mobbsie:
No problem, and yes, I agree with your decision to not pull the framework changes into the project.

On a separate note, I thought I saw somewhere that someone was working on the Abandoned Carts API.  Do you have an idea of how far along they are, and whether I can help to bring that to completion?

nozzlegear:
I think that was @alectec working on Abandoned Carts, might be a good idea to check in with him and see if there's anything you could help with!

nozzlegear:
This should now be available on Nuget with version 3.6.0. Thanks for contributing!

alectec:
@mobbsie I am working on the AbandonedCart entity and service.  I took a break but will pick it up this week.  If you want to help test, that would be great.

mobbsie:
@alectec That's great, and yes, I am keen to get involved and help. Let me know how I can contribute.

PR #169, opened 2017-08-27
Policies API

vleontyev:
Very simple Policy API


nozzlegear:
Thanks for the pull request! The Policy service should be available on Nuget in a few moments with version 4.2.0.

PR #171, opened 2017-09-13
Gift cards

vleontyev:


vleontyev:
In case if you don't have Shopify Plus. Screenshot with test results:
image

ShopifySharp/Entities/GiftCard.cs
```diff
@@ -0,0 +1,110 @@
...
+
+namespace ShopifySharp
+{
+    public class GiftCard : ShopifyObject
+    {
+        /// <summary>
+        /// Returns the amount of money remaining on the gift card.
+        /// </summary>
+        [JsonProperty("balance")]
+        public decimal Balance { get; set; }
```
nozzlegear: Change `decimal` to `decimal?`

ShopifySharp/Entities/GiftCard.cs
```diff
@@ -0,0 +1,110 @@
...
+        /// The three letter code (ISO 4217) for the currency used for the gift card.
+        /// </summary>
+        [JsonProperty("currency")]
+        public string Currency { get; set; }
+
+        /// <summary>
+        /// Returns the initial amount of money on the gift card.
+        /// </summary>
+        [JsonProperty("initial_value")]
+        public decimal InitialValue { get; set; }
```
nozzlegear: Change `decimal` to `decimal?`

vleontyev:
I've added decimal?

nozzlegear:
Cool, thanks for the pull request! It might take a day or two for me to get this published, the automated tests are failing right now and I need to dig into a bug to get it fixed.

vleontyev:
Tests failing most probably because it require Shopify Plus and they need
are activating gift card API manually by request

nozzlegear:
Oh right I forgot about that! I'll need to disable the Gift Card tests. I'm also having trouble with the fulfillment API tests though, it looks like Shopify made some kind of change that rate limits the fulfillment API even further when on a free plan (which is what I use for testing). I ran into a bug when using the smart retry execution policy, where it looks like the `HttpContent` the policy sends gets disposed after the first request and breaks further tries.

clement911:
Is the gift card API not even enabled on partner dev/test stores?

foxandrewj:
Dev stores have features equivalent to an Advanced Shopify until activated.
But the Gift Card API is a Shopify Plus feature.

clement911:
Oh I see. Thanks guys, I'm looking forward to playing the Gift Card API.

So, is it not possible to activate the Gift Card API on a dev store? Otherwise it would mean that the only way to test it out is to have an actual Plus store to test with, which seems pretty crap.

nozzlegear:
Yeah as far as I know you have to actually pay to even test the Plus APIs, which is why I've hardly implemented any (none?) of them.

clement911:
Pretty crazy. Might be worth asking if they can enable the api on a dev store... Would allow to run the tests.

nozzlegear:
Good idea, I'll get in touch with somebody at Shopify and see if we can get a Plus store for package development.

vleontyev:
But for now, can you deploy this version to nuget? Because i tested it with
my Shopify Plus store and all tests passed. I need it for our development

nozzlegear:
Sorry, I'm still working on that bug with the `SmartRetryExecutionPolicy` class which is breaking the automated tests/build. I'm working to get it fixed today and will just disable the Gift Card tests until we can get a Plus account.

clement911:
Hi @nozzlegear, just wondering if you made any progress on this one?
It just so happens that I need access to the Gift card api as well. Good timing!

nozzlegear:
Getting close! I've got what I think is a fix on my machine, just need to upload it tomorrow morning and see if it will work.

clement911:
That's good to hear!

clement911:
By the way, apparently it is possible to activate the Gift Card api on a partner store.
https://ecommerce.shopify.com/c/shopify-apis-and-technology/t/gift-card-api-access-by-public-app-469943#comment-470101

clement911:
I think that GiftCard.LineItemId should be a nullable long.
Why is it a string?

vleontyev:
Yes, correct.That was my mistake initially

nozzlegear:
Thanks for pointing that out, I'll get it changed before publish.

nozzlegear:
Okay, the Gift Card service is now available on Nuget in version 4.3.0, along with the Price Rules service. Thanks for the PR @vleontyev! 

@clement911 Thanks for the link regarding Plus accounts. I'm going to send an email to my Shopify Partner liason and see if we can get a Plus account turned on for testing ShopifySharp.

clement911:
Thanks to both of you! I will try it out.

PR #172, opened 2017-09-20
Users API

clement911:
This PR implements the user API. https://help.shopify.com/api/reference/user
It is one of the simplest API.
I wrote a couple of tests but unfortunately I'm not able to run them because I don't have access to a Shopify Plus account.

@vleontyev  any chance you could run the tests?


ernestogutierrez:
@clement911 if you are a Shopify Partner I think you can use a development store to use your tests with the Shopify Plus API. 

Have you tried that?

clement911:
l did try that but it doesn't work, even if creating a private app on the dev store. Shopify returns 404. I was told it needs to be enabled by Shopify, even on test stores! I asked them but I haven't heard back yet...

nozzlegear:
I've had the same experience. The tests for this package on AppVeyor are using a private app, but still don't have access to the Plus APIs 🤷

nozzlegear:
Okay, this should be available on Nuget with version 4.4.0. Thanks for the PR @clement911!

clement911:
Thanks!

clement911:
FYI someone at Shopify enabled the Gift Card and User APIs on my partner store and it is working nicely!

PR #183, opened 2017-10-25
Added support for Carrier Service

BrianBarnard:
Added support for the Carrier Service.


BrianBarnard:
Forgot to mention something. Shopify only seems to let you add one custom Carrier. If you try to add two you get an exception. That is why the tests look a bit different than the other services. I could not have a Create() in the initialization of the text fixture, as then any other creates would fail. I also had to add deletes after every create in each individual test, otherwise when two tests ran after each other before the Disposal the second create would fail.

nozzlegear:
Cool, thank you very much for your work on this! It's available on Nuget in version 4.6.0.

PR #185, opened 2017-10-30
Dynamic leaky bucket resize

clement911:
The SmartRetryExecutionPolicy will now dynamically resize the leacky bucket size (useful for Shopify Plus customers, who have a bigger bucket)


clement911:
For Shopify Plus customers, this should allow for much faster request processing.

clement911:
Thanks!

PR #206, opened 2018-01-18
added support for CustomerSavedSearchService, including querying for Customers in a Saved Search

pianomanjh:


nozzlegear:
Very cool, thank you for the pull request! This should be available on nuget in version 4.9.0.

Issue #209, opened 2018-01-18
Wrong JsonProperty name for Checkout AbandonedCheckoutUrl


datvm:
Currently, it is `[JsonProperty("abandoned_check_url")]` while the correct one should be `[JsonProperty("abandoned_checkout_url")]`

Source: https://help.shopify.com/api/reference/abandoned_checkouts

PR #210, opened 2018-01-18
Fixed incorrect AbandonedCheckoutUrl JsonProperty name (issue #209).

datvm:
Fix issue #209


nozzlegear:
Good catch with this! Fix should be available on Nuget in version 4.9.1

PR #213, opened 2018-01-23
Metafields service update

vleontyev:
Now service supports endpoint with parent and children ids. For example:

- /admin/products/#{id}/variants/#{id}/metafields.json
- /admin/blogs/#{id}/articles/#{id}/metafields.json


nozzlegear:
Happy to see this get merged in! I've been getting a ton of emails about using metafields and I can't wait to point them to this. Should be available on nuget in version 4.9.2!

PR #221, opened 2018-03-01
Changed the order processed at filters to DateTimeOffset

alitas:


nozzlegear:
Thanks for the pull request! I admit it would make more sense to use `DateTimeOffset` here, but I'm hesitant to change this because it would be a breaking change which I *try* to avoid in minor releases. Let me think about this and get back to you. Feel free to ping me in a day or two if I still haven't gotten back.

nozzlegear:
Okay, I think I'm going to close this PR for now. Not because it isn't useful -- we certainly should be using `DateTimeOffset` here -- but because it could break builds when people update the package. I *try* to avoid those breaking changes as much as possible, so once we get to a 5.0 release we'll revisit this and make sure it gets merged in. 

Thanks for the pull request!

Issue #225, opened 2018-03-14
Fix casing on 'MetaField'


MisinformedDNA:
'MetaField' and 'MetaFieldService' should be 'Metafield' and 'MetafieldService'. This reflects Shopify's [documenation](https://help.shopify.com/api/reference/metafield) and the .NET Design Guidelines. It's a small, but breaking change. (Added just in case the stars align.)

nozzlegear:
This is a good suggestion, but changing the name of a class would break builds, so I'm going to hold off on it until we start gearing up for a 5.0 release.

PR #232, opened 2018-03-31
Fix casing on 'MetaField'

vleontyev:
Based on issue #225 
This change will break backwards compatibility and require changes from developers after upgrade


nozzlegear:
Thanks for the pull request! Since this would be a breaking change, I'm going to delay this one until we start working on a 5.0 release.

vleontyev:
@nozzlegear  Let me know if you need my help with 5.0 release

PR #249, opened 2018-05-21
DraftOrderService

nozzlegear:
This adds support for the Draft Orders API: https://help.shopify.com/api/reference/orders/draftorder


nozzlegear:
Can't wait to use this one, it looks like it opens up some really interesting opportunities!

PR #250, opened 2018-05-21
Added FulfillmentServiceService

davron12:
Adds support for Shopify's FulfillmentService API


ShopifySharp/Filters/ListFilter.cs
```diff
@@ -40,5 +40,13 @@ public class ListFilter : CountFilter
         /// </summary>
         [JsonProperty("order")]
         public string Order { get; set; }
+
+        /// <summary>
+        /// An optional field name to include fulfillment services from entire store.
+        /// Any valid values are null and "all"
+        /// </summary>
+        [JsonProperty("scope")]
+        public string Scope { get; set; }
```
nozzlegear: Since the `ListFilter` class is used and inherited in several different places, could you create a new class named `FulfillmentServiceListFilter` that inherits from this `ListFilter` and adds the `Scope` property? That way we don't have the scope property showing up for unrelated services that also use this `ListFilter`.

It would look something like this:

```cs
public class FulfillmentServicesFilter : ListFilter 
{
    /// summary here
    [JsonProperty("scope")]
    public string Scope { get; set; }
}
```

ShopifySharp/Services/FulfillmentService/FulfillmentServiceService.cs
```diff
@@ -0,0 +1,106 @@
...
+        /// <param name="myShopifyUrl">The shop's *.myshopify.com URL.</param>
+        /// <param name="shopAccessToken">An API access token for the shop.</param>
+        public FulfillmentServiceService(string myShopifyUrl, string shopAccessToken) : base(myShopifyUrl, shopAccessToken) { }
+
+        /// <summary>
+        /// Gets a list of your app's FulfillmentServices.
+        /// </summary>
+        /// <param name="options">Options for filtering the list. Set scope to all to retrieve all of the store's fulfillment services</param>
+        /// <returns>The list of fulfillment services matching the filter.</returns>
+        public virtual async Task<IEnumerable<FulfillmentServiceEntity>> ListAsync(ListFilter filter = null)
```
nozzlegear: Change `ListFilter filter = null` to `FulfillmentServicesFilter filter = null` once that class has been created.

ShopifySharp/Entities/FulfillmentServiceEntity.cs
```diff
@@ -0,0 +1,88 @@
+using Newtonsoft.Json;
+using System;
+using System.Collections.Generic;
+
+namespace ShopifySharp
+{
+    /// <summary>
+    /// An object representing a Shopify fulfillment service.
+    /// </summary>
+    public class FulfillmentServiceEntity : ShopifyObject
```
nozzlegear: Rename this class to `FulfillmentService` rather than `FulfillmentServiceEntity`.

ShopifySharp/Services/FulfillmentService/FulfillmentServiceService.cs
```diff
@@ -0,0 +1,106 @@
...
+using System.Collections.Generic;
+using System.Threading.Tasks;
+using ShopifySharp.Infrastructure;
+
+namespace ShopifySharp
+{
+    /// <summary>
+    /// A service for manipulating Shopify fulfillments.
+    /// </summary>
+    public class FulfillmentServiceService : ShopifyService
```
nozzlegear: Just curious, I haven't looked at the docs yet. Does this endpoint support counting the Fulfillment Services?

ShopifySharp/Entities/FulfillmentServiceEntity.cs
nozzlegear: Also the file should be `FulfillmentService.cs` once renamed, sorry!

ShopifySharp/Services/FulfillmentService/FulfillmentServiceService.cs
davron12: No, there's no counting documented. I'm assuming most stores don't have a ton of fulfillment services registered, so there wasn't a reason to implement a count function.

ShopifySharp/Entities/FulfillmentServiceEntity.cs
davron12: That's originally what I wanted to name it, but because Entities and Services aren't in their own namespace, there will be a conflict with the Service already named FulfillmentService.

ShopifySharp/Filters/ListFilter.cs
davron12: Yes, only scope and none of the other parameters. I'll make it a parameter for the method then.

davron12:
I've made all the changes you requested, also fixed the spelling of Fulfillment in Fulfillment_Tests.cs, and added the OpenAsync method to the FulfillmentService.cs.

ShopifySharp/Entities/FulfillmentServiceEntity.cs
davron12: Yeah, I didn't like adding the suffix either, but didn't see any way around it.

nozzlegear:
Published in 4.14.0, thank you very much!

Issue #254, opened 2018-05-27
Missing property in Fulfillment entity


kgrvr:
As per Shopify docs here: https://help.shopify.com/api/reference/shipping_and_fulfillment/fulfillment

A warning specifies: After July 1st, 2018, apps will be required to specify a location_id when creating fulfillments.

Fulfillment entity don't contain **location_id**.

nozzlegear:
Good catch! I'll get this in one of the next releases.

davron12:
@nozzlegear - I already got this, as well as an undocumented destination property. I'll be sending a pull request soon.

PR #255, opened 2018-05-29
Added Missing Properties to Fulfillment Class

davron12:
Added destination (appears to be an Address type), email, notify_customer, variant_inventory_management, and location_id to the fulfillment class.

`destination` is the shipping address for the fulfillment.

`email` and `notify_customer` appear to work together to generate an email notification to the customer based on updates to the fulfillment.

`location_id` is to support the new Shopify multi-location fulfillments. #254


nozzlegear:
Published in 4.14.1. Thank you!

PR #264, opened 2018-06-16
PrepareQuerystring, use StringComparer.Ordinal for query ordering

yuvalsade:
support uppercase querysting parameters for request authentication.
shopify is using ruby ordering (A > a > z), c# default is  ( z > A > a), it needs to be done the same, so that requests with upper case parameters would not fail authentication.


yuvalsade:
AuthorizationService/EncodeQuery
s.Replace("%", "%25").Replace("&", "%26"), misses on some cases, changed it to Uri.UnescapeDataString

nozzlegear:
Good catch! I don't have the foggiest idea about how Ruby does anything, so I would not have known about this.

yuvalsade:
can you add me to the wall? :)

nozzlegear:
Sure!

nozzlegear:
Your changes are published on Nuget in 4.15.1. Thank you!

PR #280, opened 2018-07-26
 Added ability to use HttpRequestHeaders to AuthorizationService.IsAuthenticWebhook

dkershner6:
I don't use ASP.net, so getting it into the correct form is quite a task.  A bigger task than just adding this functionality.  The method uses the native HttpRequestHeaders instead of IEnumerable<KeyValuePair<string, StringValues>>, but is otherwise identical.


nozzlegear:
Thank you! I'm not home at the moment, but I'll get this published to nuget as soon as I can.

nozzlegear:
Sorry, this one got away from me. It's published now in 4.16.2. Thank you!

PR #302, opened 2018-11-06
Simplest Possible GraphQL Implementation

dkershner6:
I required the Graph API for a project I'm working on, and since it just went into general availability, I thought I would share with all of you.  This module was designed with this mindset:
- Maximum flexibility regarding the output of the graph API, since it can be anything.
- Minimum effect on the current REST API infrastructure.

To that end, since Graph has it's own error system that is very different from the REST calls, I separated it out, and in the process lost a bit of minor exception information (requestId).

Other than that, this should provide a fully functioning GraphService that returns a JToken on success.

### DIFF
--- a/ShopifySharp/Services/Graph/GraphService.cs
+++ b/ShopifySharp/Services/Graph/GraphService.cs
@@ -1,0 +1,71 @@
+﻿using Newtonsoft.Json.Linq;
+using System.Net.Http;
+using ShopifySharp.Filters;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using ShopifySharp.Infrastructure;
+using Newtonsoft.Json;
+using System.Net;
+
+namespace ShopifySharp
+{
+    /// <summary>
+    /// A service for using or manipulating Shopify's Graph API.
+    /// </summary>
+    public class GraphService : ShopifyService
+    {
+        /// <summary>
+        /// Creates a new instance of <see cref="GraphService" />.
+        /// </summary>
+        /// <param name="myShopifyUrl">The shop's *.myshopify.com URL.</param>
+        /// <param name="shopAccessToken">An API access token for the shop.</param>
+        public GraphService(string myShopifyUrl, string shopAccessToken) : base(myShopifyUrl, shopAccessToken) { }        
+
+        /// <summary>
+        /// Executes a Graph API Call.
+        /// </summary>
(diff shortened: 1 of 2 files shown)


nozzlegear:
Very cool, thank you!! I'm out of the office until tomorrow so I'll review and merge it then. Can I ask what you needed the graphql API for that you couldn't do with rest? Just curious, I found myself recently wishing I could easily is graphql API for a client's project.

dkershner6:
Couldn't isn't my feeling, but COULD do 500 calls per 20 seconds rather than 40?  That I like.  It handles metafields in a much stronger way than the REST one does as well.

REST is still absolutely critical with any call that includes BodyHtml though.

They are similar tools, but with very different "cost" structures.  I like having access to both.

clement911:
Nice one, I'll definitely give this a try!

nozzlegear:
@dkershner6 would you mind adding a quick usage example to the readme before I merge this?

dkershner6:
I wouldn't want to presume where to put it...but I will put a little blurb here and you can drop it where it makes the most sense.

dkershner6:
**Graph API Calls**

You can make customized API calls to Shopify's Graph API using GraphService to handle all of the authentication and error handling for you.

Example Request for all the collections that a product belongs to (returning just one):
```cs
Product product = new Product() { Id = 12345 }; //Any Product

GraphService graphService = new GraphService(shop, token);

JToken response = await graphService.PostAsync("query { product (id: \"gid://shopify/Product/" + product.Id + "\") { collections (first: 1) { edges { node { title } } } } }");  //Any viable query
```

Response (inside the JToken):
```json
"products": {
            "edges": [
                {
                    "node": {
                        "collections": {
                            "edges": [
                                {
                                    "node": {
                                        "title": "Test Collection"
                                    }
                                }
                            ]
                        }
                    }
                }
            ]
        }
```

dkershner6:
I added a target framework of .NET Core 2.1 as well.  I use ShopifySharp on it almost exclusively now, and it has worked without a hitch.

Open to any reasons why this isn't a good idea, but figured it was worth doing.

dkershner6:
I also have another pull request ready after this one.  I feel that this one is a no-brainer in terms of adding functionality at a basic and simple level, but the next is a little more of a proposal in nature, so I don't want to combine them.

nozzlegear:
Oh shoot, I let this get away from me. Sorry about that! I don't have any issues with the new .NET Core target.

nozzlegear:
Published in 4.17.0. Thank you! I'm excited to try the graph API out myself.

PR #305, opened 2018-11-18
Proposal: GraphExtensions

dkershner6:
Since the Graph API is a lot more flexible on how it responds, and the possible permutations on what people may do is vast, the simple implementation and flexible response was the right way to start.

But, there are some very powerful and likely to be common uses of the Graph API that I feel warrant their own extension methods.

The format for these, and exactly how to implement them is definitely debatable, but here is an example of how I implemented Inventory calls for one of my apps.  Very open to recommendations, but these two methods and 3 entities are tested as working and are about to head into production.

### DIFF
--- a/ShopifySharp/Entities/GraphKeyValuePair.cs
+++ b/ShopifySharp/Entities/GraphKeyValuePair.cs
@@ -1,0 +1,19 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace ShopifySharp
+{
+    public class GraphKeyValuePair
+    {
+        /// <summary>
+        /// In the format gid://...
+        /// </summary>
+        public string Id { get; set; }
+
+        /// <summary>
+        /// Will be the most important value for whatever you are calling, i.e. the metafield value in a metafield call.
+        /// </summary>
+        public object Value { get; set; }
+    }
+}
--- a/ShopifySharp/Entities/GraphMutationResponse.cs
+++ b/ShopifySharp/Entities/GraphMutationResponse.cs
@@ -1,0 +1,19 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace ShopifySharp
+{
+    public class GraphMutationResponse
+    {
+        /// <summary>
+        /// List of returned values with the mutation call.
+        /// </summary>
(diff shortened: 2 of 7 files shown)


dkershner6:
For those not as familiar with the Graph API in Shopify, the combination of these two calls uses about 10 Graph cost.  You receive 50 per second, so this would allow you to run 5 updates per second rather than REST, which would require 1-2 seconds per update, depending on whether you have the inventory item ID or not.  A gain of 500-1000% throughput.

clement911:
Thanks for that.
I still haven't had a chance to play with GraphQL so it's not obvious to me what patterns are common.
Just one thing, are you not worried that if many people contributes their own extensions/queries into the library, then it's going contain a huge amount of code that only serves very specific queries?

dkershner6:
I think it fits closely enough content wise to be in the same project.

Pollution? Maybe, but i'd call it highly likely that Shopify makes the Graph API a first class citizen from here forward, and the REST API a second class one, just based on their own docs (They had a migration guide from day 1 public).  So, we will need to eventually figure something out, most likely.

Regarding pollution, I am actually finding the mutations to be easier to handle response wise than the queries (which can be anything).  In addition, the mutations are tougher to code.  I'd say there is definite value on that side, and nearly every response can be handled by my MutationResponse entity.

Querying is easier, but a tougher call... maybe just a minimum standard for what qualifies as an extension, not sure.

clement911:
To clarify, I definitely think the GraphQL service should be in the core lib, my comment was about baked in queries.

dkershner6:
Agree, but I also think that the Graph API is less intuitive than the REST one, so the help (baked in) might be even more necessary.  Leaving the Graph API with having to type the full, difficult to manage, no intellisense, no formatting body, will likely make it quite cumbersome to use.

Hell, I built the service and I am building myself extensions so I don't have to use it (besides when I build the extension anyway).

clement911:
Got you.
Thanks again for sharing your implementation and I'm looking forward to playing with it!

dkershner6:
After having created a bunch of these, I agree that this organization structure could get pretty hairy.  Intellisense is getting pretty long for me already, need a way to partition them.

I will think on this some more.

PR #318, opened 2018-12-12
Minor refactoring of the GraphService along with a simple JSON (JToken) overload for the potential inclusion of JSON variable objects to the calls.

dkershner6:


nozzlegear:
Published in 4.17.3!

PR #323, opened 2019-01-09
Add support to AdjustInventoryLevel in InventoryLevelService

arsuceno:
There is no support for the adjust.json endpoint within inventory_levels: https://help.shopify.com/en/api/reference/inventory/inventorylevel#adjust . This PR is to give support for that endpoint.


nozzlegear:
Published in 4.18.0, thank you!

PR #332, opened 2019-02-07
Adds support for the TaxCode attribute for Product Variants

joeybeninghove:
This is the tax_code field used in Shopify for Product Variants for Shopify Plus stores that have the AvaTax app installed for calculating tax rates.


nozzlegear:
Thank you! Sorry this took so long to get merged. I had to revert the test additions you made since the TaxCode property is only available on shops that have a special tax app installed, but the `ProductVariant.TaxCode` property has now been published in 4.18.2 on Nuget.

PR #342, opened 2019-03-07
Updated ProductVariant ListAsync method, to support listing variants for all products

samnorr:
Noticed that there was no ability to get variants for ALL products, despite the Shopify API allowing this (although it seems to be undocumented).

### DIFF
--- a/ShopifySharp/Services/ProductVariant/ProductVariantService.cs
+++ b/ShopifySharp/Services/ProductVariant/ProductVariantService.cs
@@ -30,13 +30,15 @@ namespace ShopifySharp
         }
 
         /// <summary>
-        /// Gets a list of variants belonging to the given product.
+        /// Gets a list of variants, optionally filtered to a given product.
         /// </summary>
         /// <param name="productId">The product that the variants belong to.</param>
         /// <param name="filterOptions">Options for filtering the result.</param>
-        public virtual async Task<IEnumerable<ProductVariant>> ListAsync(long productId, ListFilter filterOptions = null)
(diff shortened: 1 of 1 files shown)


clement911:
I didn't this worked? This is not documented anywhere on https://help.shopify.com/en/api/reference/products/product-variant

dkershner6:
It actually does work, I just tried it.  I can't think of many use cases, and I think support might someday just disappear, but it does seem to currently work.

samnorr:
Yeah it is undocumented, but definitely works.

The main use case for me would be to get variant properties (such as inventory_item_id), for all variants. This is a bit simpler than the alternative of having to download the full Products list, and then extracting and merging the variants from that.

clement911:
I see.
Although I think loading products would be a bit faster since you can page 250 products at a time.

nozzlegear:
Thanks for the pull request! I do appreciate the effort you put into it, but I'm wary of adding features into the package that aren't officially documented or supported by Shopify -- if they break down the road we'll get developers looking for answers on why their own apps using ShopifySharp have stopped working. 

That said, your use-case (getting variant properties for all variants) is obviously a valid one. My suggestion would be to take the ShopifySharp.ProductVariantService and extend it with this method: 

```cs
public class CustomVariantService : ShopifySharp.ProductVariantService
{
    public CustomVariantService(string domain, string accessToken) : base(domain, accessToken)
    {
    }

    public async Task<IEnumerable<ProductVariant>> ListAllVariantsAsync()
    {
        ....
    }
}
```

That way you (and anyone reading this in the future) can still have this functionality, but we as package authors don't have to officially support this (until Shopify themselves start documenting it).

Again, thank you for the pull request! I hope this doesn't discourage you from contributing again in the future, I hate having to turn down contributions.

PR #352, opened 2019-04-09
Shopify now also returning 422 errors on too many requests

mgranell:
In testing, Shopify provided a 422 error with "Too many requests" occasionally, rather than the expected 429 request. 

In this pull requests I've limited the retry logic only when the phrase "Too many requests" is present, which is fragile, but I didn't want to do all 422 responses.

### DIFF
--- a/ShopifySharp/Services/ShopifyService.cs
+++ b/ShopifySharp/Services/ShopifyService.cs
@@ -212,7 +212,8 @@ namespace ShopifySharp
             var code = response.StatusCode;
 
             // If the error was caused by reaching the API rate limit, throw a rate limit exception.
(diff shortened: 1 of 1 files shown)


clement911:
Under what scenario does this occur?
Is that when the newly introduced limits are breached? https://help.shopify.com/en/api/getting-started/api-call-limit
It might be a good idea to post in the Shopify API forums as this sounds like a bug to me

mgranell:
I haven't seen it since in testing (although I had the retry policy on). This was a standard API request (not a GraphQL one). Happy to close this pull request for now and if the error comes up again ask in the forum.

PR #353, opened 2019-04-09
Change JSON Parsing to not lose time zone information 

mgranell:
When querying the GraphQL endpoint with Shopify, the dates come base with time zone information. However the default JToken.Parse() strips the timezone information for dates and returns them as DateTime instances. This makes any conversion to DateTimeOffset incorrect as it picks up the local time.

Changing to explicitly not handling datetimes in the _tokenising_ part of the parsing means that date times are correctly converted to DateTimeOffset when de-serializing.

### DIFF
--- a/ShopifySharp.Tests/DateTime_Tests.cs
+++ b/ShopifySharp.Tests/DateTime_Tests.cs
@@ -1,0 +1,79 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace ShopifySharp.Tests
+{
+    using System.Diagnostics;
+    using System.Globalization;
+    using System.Linq;
+    using System.Net;
+    using System.Threading.Tasks;
+
+    using Newtonsoft.Json;
+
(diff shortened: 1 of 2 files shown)


clement911:
Doesn't this make dates be returned as strings?

dkershner6:
I don't think so, I think he means some json serialization changes. I was just too lazy or forgot when I built it, unlike with the REST API.  Json dates are in fact strings, it's Json.net making em dates.

@mgrannel , you happen to have the jsonserializer setting in question on hand?

mgranell:
I've just added a test that demonstrates the issue. There is a problem that if a DateTime JToken has come from a JSON string with a timezone (or UTC), when serialized as string it loses that information. This tripped me up when using the library as I expected the datetime to come through cleanly, however when parsed into a datetimeoffset the timezone was wrong.

mgranell:
The other option is to try and ensure people never de-serialize the datetimes into string fields/properties.

PR #355, opened 2019-04-10
API versioning

clement911:
This is a non breaking change since no default version is set.
The requested urls will remain the same unless the user explicitly sets an API version.

### DIFF
--- a/ShopifySharp/Entities/ShopifyObject.cs
+++ b/ShopifySharp/Entities/ShopifyObject.cs
@@ -21,5 +21,8 @@ namespace ShopifySharp
         /// </remarks>
         [JsonProperty("id")]
         public long? Id { get; set; }
+
+        [JsonProperty("admin_graphql_api_id")]
+        public string AdminGraphQLAPIId { get; set; }
     }
 }
--- a/ShopifySharp/Services/AccessScope/AccessScopeService.cs
+++ b/ShopifySharp/Services/AccessScope/AccessScopeService.cs
@@ -16,9 +16,7 @@ namespace ShopifySharp
         /// </summary>
         /// <param name="myShopifyUrl">The shop's *.myshopify.com URL.</param>
         /// <param name="shopAccessToken">An API access token for the shop.</param>
-        public AccessScopeService(string myShopifyUrl, string shopAccessToken) : base(myShopifyUrl, shopAccessToken)
-        {
-        }
+        public AccessScopeService(string myShopifyUrl, string shopAccessToken) : base(myShopifyUrl, shopAccessToken) { }
 
         //oauth endpoints don't support versioning
         protected override bool SupportsAPIVersioning => false;
--- a/ShopifySharp/Services/AccessScope/AccessScopeService.cs
+++ b/ShopifySharp/Services/AccessScope/AccessScopeService.cs
@@ -18,9 +18,10 @@ namespace ShopifySharp
         /// <param name="shopAccessToken">An API access token for the shop.</param>
         public AccessScopeService(string myShopifyUrl, string shopAccessToken) : base(myShopifyUrl, shopAccessToken)
         {
-            //outh endpoints don't support versioning
-            SetAPIVersion(null);
         }
+
+        //oauth endpoints don't support versioning
+        protected override bool SupportsAPIVersioning => false;
 
         /// <summary>
         /// Retrieves a list of access scopes associated to the access token.
--- a/ShopifySharp/Services/AccessScope/AccessScopeService.cs
+++ b/ShopifySharp/Services/AccessScope/AccessScopeService.cs
@@ -16,7 +16,11 @@ namespace ShopifySharp
         /// </summary>
         /// <param name="myShopifyUrl">The shop's *.myshopify.com URL.</param>
         /// <param name="shopAccessToken">An API access token for the shop.</param>
-        public AccessScopeService(string myShopifyUrl, string shopAccessToken) : base(myShopifyUrl, shopAccessToken) { }
+        public AccessScopeService(string myShopifyUrl, string shopAccessToken) : base(myShopifyUrl, shopAccessToken)
+        {
+            //outh endpoints don't support versioning
+            SetAPIVersion(null);
+        }
 
         /// <summary>
         /// Retrieves a list of access scopes associated to the access token.
--- a/ShopifySharp/Services/ShopifyService.cs
+++ b/ShopifySharp/Services/ShopifyService.cs
@@ -14,7 +14,7 @@ namespace ShopifySharp
 {
     public abstract class ShopifyService
     {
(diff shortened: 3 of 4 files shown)


dkershner6:
I had the same idea. Thumbs up from me.

aztekweb:
Thanks for getting to this so quickly.  I read the Shopify blog post, came here to open an issue, and found this PR already done.

nozzlegear:
Thanks @clement911! I've got some free time on Friday and will be going through this and my other OS projects. I want to play around with this before I merge it, I've been pondering a different way of handling the API versioning and will explain Friday. I'm glad they're _finally_ doing this! 🎉

clement911:
@nozzlegear, I'd love to see what you're thinking of.
There is no rush to merge this in...

clement911:
Hi @nozzlegear ,
Are you in a position to share you thinking around handling API versioning?

nozzlegear:
Yes! Sorry, I've been so busy lately it's crazy. 😬 To give you a basic idea of what I'm thinking about, it might be nice as maintainers to only officially support one version of the Shopify API, [something they do with the Stripe.Net package](https://github.com/stripe/stripe-dotnet/blob/83ecbac379eb291c688a13faf69e9976ae28b7fd/src/Stripe.net/Infrastructure/Public/StripeConfiguration.cs#L12) (the package I had drew inspiration from when I wrote the original version). Pinning the package to one specific API version means all of the objects/calls are known to work for that version, and we don't have to do anything funky to make things work with multiple versions. That doesn't mean we just sit on one version forever, it can be changed as new versions release.

clement911:
Interesting. This certainly has the benefit of being simple and easy to maintain.
So you're saying that if a new version is released by Shopify, then we simply update the version string in ShopifySharp and start implementing the new features?
In other words, if someone needs to target an old version, they would need to use a previous version of the nuget package, right?
I like how simple this is even though it removes flexibility.
Personally, I don't foresee a need to use different versions in a single app...

clement911:
Looking at this PR, it might just be a matter of making SetGlobalAPIVersion private and called internally with the latest version on startup.

Also need to remove the SetAPIVersion override method (although I suppose it doesn't hurt to leave it if for some reason someone needs)

What do you think?

clement911:
@nozzlegear  @dkershner6 
I've simplified the PR to pin the package to a single static version.
What do you think?

clement911:
It's pretty simple but I haven't tested yet. Just wanted to get your thoughts..

nozzlegear:
I like that a lot! That's just about exactly what I was thinking. Could we make the version virtual, so that if people _really, really_ want to use a different version they can extend a service and set it themselves?

clement911:
All done :)

clement911:
I was also thinking if moving forward we should have two branches / nuget package.

The 'master' branch for the stable API version, released as a stable nuget package.

A new 'rc' branch for the upcoming release candidate version, released as a preview nuget package.

See https://help.shopify.com/en/api/versioning

The reason is that we are going to want to implement and test new features of RC versions before they become stable.

Once a RC becomes stable, we can merge 'rc' into 'master' and use 'rc' for the next release candidate version.

What do you think of that approach?

clement911:
Thinking about more about this, I think it would also be a good idea to have a logical versioning number scheme that matches the API version.

Example: if a release is targeting version 20**19-04**, it should have a nuget version of **1904.[minor].[patch]** (the major must be less than 65535 so it cannot be 201804.19.3).

So to get started we would have:
**master => 1904.[minor].[patch]**
**rc => 1907.[minor].[patch]-rc**

This way, you can set up your build to only accept upgrades that target the same major version, but upgrading to the next API version must be done explicitely.

What do you think?

@nozzlegear I have no idea how much work is involved to change your current nuget release process to add an additional rc branch?

clement911:
I've also just added an new property 'AdminGraphQLAPIId' to ShopifyObject.
I'm finally starting to look into GraphQL and this property should be quite useful.
See https://help.shopify.com/en/api/graphql-admin-api/migrating-from-rest

@nozzlegear any objection to release this PR into a nuget package (we can worry about the new version scheme a bit later I guess).

dkershner6:
The proposed versioning based on shopify's APIs feels like it prevents us from ever having a breaking change on the ShopifySharp side (unless it somehow matches a breaking change on Shopify's side).

I would also argue that the changes between versions are likely to be small.  If anyone has any experience with twitter or facebook, additions are usually just additive, and therefore you can move between version numbers mostly insignificantly.

I'd say we just declare that we are always on the most recent stable release and keep up with the same versioning as now.

The rc branch seems like a decent idea.  I think adding new properties to master, while not breaking anything, would be confusing for someone who couldnt yet use them without a special opt in (although I guess we could handle through warnings, this feature requires setting version to...).  The branch would allow for breaking stuff though.

clement911:
> The proposed versioning based on shopify's APIs feels like it prevents us from ever having a breaking change on the ShopifySharp side (unless it somehow matches a breaking change on Shopify's side).

That's a good point.
I don't think Shopify itself will make a breaking change in any given version unless there is a security risk.
By definition released versions are stable.
But still, we might make to introduce a breaking change in a given ShopifySharp version, which might be a problem.
We could append an additional digits for breaking changes.
19040.[minor].[patch]
19041.[minor].[patch]
19042.[minor].[patch]
etc...
Is that getting too complicated?

> I would also argue that the changes between versions are likely to be small. If anyone has any experience with twitter or facebook, additions are usually just additive, and therefore you can move between version numbers mostly insignificantly.

I'm not sure about that.
I think the main reason for the versioning is that they can introduce breaking changes if they need to.

> I'd say we just declare that we are always on the most recent stable release and keep up with the same versioning as now.

That could certainly work as well.

> The rc branch seems like a decent idea. I think adding new properties to master, while not breaking anything, would be confusing for someone who couldnt yet use them without a special opt in (although I guess we could handle through warnings, this feature requires setting version to...). The branch would allow for breaking stuff though.

I do believe they might introduce breaking changes in future versions so I feel pretty strongly we need an RC branch.

nozzlegear:
I tend to agree that introducing the extra versions is starting to get a little complicated, but it is also useful. I think a compromise (for now) might be to call out which version is officially supported by the package, probably somewhere in the readme. We can also maintain a list of which ShopifySharp versions supported which Shopify version:

```
4.19.x => 2019-04
4.x.x => 2019-xx
4.x.x => 2020-xx 
...
```

It's not perfect of course, I'm absolutely willing to entertain other suggestions!

clement911:
Sure. let's start simple.

nozzlegear:
Just a quick update, I've got this merged and I'm adding in some fixes/changes for other issues. New release will be in a couple days;

clement911:
Sounds good.

nozzlegear:
Released in 4.20.0!

PR #368, opened 2019-05-30
Added optional count filter to customers/count.json

clement911:


nozzlegear:
Available in 4.20.0 👍

Issue #380, opened 2019-07-07
2019-07 breaking change for pagination


clement911:
It looks like the `page` parameter is not supported on some endpoints in version 2019-07.

See https://help.shopify.com/en/api/guides/paginated-rest-results

Might need to extract the new `Link` header and deprecate the `page` param for the affected endpoints.

nozzlegear:
Hmm that's an interesting change. So instead of returning the list of items, we'd have to return a new object with both the list and a property pointing to the next "link". 

```cs
class ListResult<T>
{
    public IEnumerable<T> Items { get; set; }
    public string NextPageLink { get; set; }
}
```

clement911:
Yes, it's similar to cursor-based paging with GraphQL.
This change will be touching quite a few files in ShopifySharp I'm afraid...

mattxo:
Hi @clement911 

Thanks for those references on #382. That certainly complicates things!

I can only speak for myself but it would be a triviality for me to implement IListAsync across all services. I could do a PR for that today no problem. 

As a balance of effort vs reward, for me personally it's absolutely worth it, but how it would affect the project more broadly? I'm not sure... 

That said, 2019-04 is supported until 2020-04 and only 7 services will be modified to use cursor-based pagination in 2019-07. 

Orders, most notably, is absent from the list of services that will get cursor-based pagination in 2019-07 which inclines me towards thinking an implementation of IListAsync which covers the rest of the services may quite useful for quite some time.

Specifically in regards to implementing cursor based pagination, my initial thoughts are that one would remove ListAsync() from the relevant services and add PageAsync() (or perhaps more specifically PageAsync.

Something like this perhaps?

```
public class PaginatedRequest 
{
    public string PageInfo { get; set; }
    public int? Limit { get; set; }        
}

public class PaginatedResponse<T>
{
    public IEnumerable<T> Items { get; }        
    public string Link { get; }
    public PaginatedRequest NextPage { get; }
    public PaginatedRequest PreviousPage { get; }
}

public class ProductService : ShopifyService
{
    public ProductService(string myShopifyUrl, string shopAccessToken) : base(myShopifyUrl, shopAccessToken) { }

    public async Task<PaginatedResponse<Product>> PageAsync(PaginatedRequest request)
    {
        var result = new PaginatedResponse<Product>(); // Do magic

        return await Task.FromResult(result);
    }
}
```

clement911:
[...]

PR #382, opened 2019-07-10
Add GetMetaFieldsAsync() to OrderService

mattxo:


nozzlegear:
Hey @mattxo, thanks for the pull request! It looks like there's a lot going on here beyond just adding metafields to the OrderService, so I'd like to break this up into multiple different PRs just to keep everything focused and have a clear reference to which PR did what if we ever need to come back to review things in the future. If you don't mind, you could break up your changes into different git branches and submit them for these things:

- The metafield changes. 
- The inventory level service. 

I see you've made a few other changes, like adding extension methods for converting objects to JSON, and extensions/interfaces for listing objects in a loop. I appreciate the effort you've put in and hate to have it go to waste, but I think for now I want to keep extension methods out of the package so I don't get in trouble for some subtle bug or breaking issue if developers start to rely on them and something changes. I'm already planning to remove the existing `ToDictionary()` extension method in a future release. 

Thanks again for the PR, and please let me know if you need any help splitting this up into smaller pull requests!

clement911:
I don't want to be a downer either but I agree very much with @nozzlegear 
I think this project values having good consistency and having a very thin layer on top of the Shopify API. Some of these additions do more than necessary. It's not necessarily wasted but you would probably put these extensions in your own lib rather that directly into ShopifySharp.

mattxo:
Hi @nozzlegear and @clement911

Thanks for the feedback and no worries at all! 

I'll break down those two you mentioned into seperate PR's :)

How do you feel about including the IListAsync interface (without including an implementation in ShopifySharp?) 

Ie, 

```
public interface IListAsync<TShopifyObject, TListFilter>
    where TShopifyObject : ShopifyObject
    where TListFilter : ListFilter
{
    Task<IEnumerable<TShopifyObject>> ListAsync(TListFilter options = null);
}
```

and 

```
public class ProductService : ShopifyService, IListAsync<Product, ProductFilter>
```

An IListAsync interface (or similar) is necessary for a 3rd party library to deal with the need to get all available entities for all every type of entity without breaking DRY badly. 

I'd be more than happy to create/contribute an extensions library, eg. ShopifySharp.Extensions or similar but it would be much neater if ShopifySharp could expose those extension points.

nozzlegear:
I'm interested in adding interfaces to this package for the services, it's something that I want to get in relatively soon as I've got a personal need for it. Adding interfaces for services that list objects would be cool, but I need some additional time to review and make sure it will work. I can't remember for sure off the top of my head, but I think there are a couple of services that wouldn't work with a list interface (e.g. ones that require a parent id in addition to an optional filter).

As for extensions, I'll see about opening up a branch for a new package soon! There's some experimental F# stuff I want to play around with too, so I might even do a ShopifySharp.Experimental or package.

mattxo:
More than happy to contribute as I have a rather pressing need for this too :) 

A quick review on my part suggests that you're perfectly correct in suggesting some services that implement a ListAsync() method are not compatible with the interface I suggested above. Not only is it that some require a parent Id, but that some either do not return a collection of ShopifyObject's and/or their filter parameter does not derive from ListFilter (rather, Parameterizable). 

Further, there are a variety method signatures, eg.

```
ListAsync()
ListAsync(string fields = null)
ListAsync(long parentId)
ListAsync(long parentId, Parameterizable filter = null)
ListAsync(long parentId, string fields = null)
ListAsync(long? sinceId = null, string handle = null, string fields = null)
ListAsync(long? sinceId = null, string fields = null)
ListAsync(Parameterizable options = null)
ListAsync(long orderId, long fulfillmentId)
ListAsync(long productId, long? sinceId = null, string fields = null)
ListAsync(long? sinceId = null, string fields = null)
ListAsync(long orderId, long? sinceId = null, bool? inShopCurrency = null)
```

These however, can boil down to

```
ListAsync()
ListAsync(Parameterizable options)
ListAsync(long parentId)
ListAsync(long parentId, Parameterizable options)
```

By forwarding the optional parameters 

eg.

```
public virtual async Task<IEnumerable<ApplicationCredit>> ListAsync()
{
    return await ListAsync(null);
}
```

Which suggests that perhaps...

```
public interface IListAsync<TShopifyObject, TListFilter>
    where TShopifyObject : ShopifyObject
    where TListFilter : Parameterizable
{        
    Task<IEnumerable<TShopifyObject>> ListAsync(TListFilter options = null);
}

public interface IListChildrenAsync<TShopifyObject, TListFilter>
    where TShopifyObject : ShopifyObject
    where TListFilter : Parameterizable
{        
    Task<IEnumerable<TShopifyObject>> ListAsync(long parentId, TListFilter options = null);
}

public interface IListAsync<TObject>        
{
    Task<IEnumerable<TObject>> ListAsync();
}

public interface IListChildrenAsync<TObject>
{
    Task<IEnumerable<TObject>> ListAsync(long parentId);
}
```

... would be suitable. It's also possible to boil the 4 interfaces down to 2 depending on what generic constraints are required (if any), which looks more like 

```
public interface IListAsync<TResult, TOptions> where TOptions : Parameterizable        
{
    Task<IEnumerable<TResult>> ListAsync();
    Task<IEnumerable<TOptions>> ListAsync(TOptions options = null);
}

public interface IListChildrenAsync<TResult, TOptions> where TOptions : Parameterizable
{
    Task<IEnumerable<TResult>> ListAsync(long parentId);
    Task<IEnumerable<TResult>> ListAsync(long parentId, TOptions options = null);
}
```

I think making ```TOptions options = null``` just ```TOptions options``` would be cleaner and I don't *think* that would change the existing API... (that said, I have essentially zero experience contributing to public libraries so... I could be completely wrong!)

I went through and implemented the first dozen services according to the above interfaces and no changes needed to be made to any class aside from adding the interface and a forwarding method where necessary. 

Happy to submit a PR to this effect for review/discussion if it would be of use?

clement911:
Hi @mattxo 

I like your idea of bringing some consistencies across the different services with interfaces.
Before settling on the final design, I want to point out that the latest version of the REST API 2019-07 introduces a significant breaking change in the way pagination works.

See https://github.com/nozzlegear/ShopifySharp/issues/380
See https://help.shopify.com/en/api/guides/paginated-rest-results
See https://help.shopify.com/en/api/versioning/release-notes/2019-07

ShopifySharp is currently using 2019-04 but the intent is to keep it up to date with the latest stable version, which means we'll have to implement the new cursor pagination.
This pagination is another reason why it makes even more sense to have some consistency in the way ListAsync<> works across different services, because the pagination mechanism will work the same.

I think the new design make take cursor pagination into consideration.
It will be a more work but if we don't do it now, your implementation will have to change almost immediately so I think it would be worth it.

Specifically, the `Link` http header must be extracted and returned along with the page so that the consumer can fetch the next/previous page.
Instead of returning a `Task<IEnumerable<TResult>>`, it should probably return a `Task<Page<TResult>>`, with the page class being something similar to this:
```
public class Page<T>
{
     IEnumerable<TResult> Results {get;}
     string NextPageCursor {get;}
     int Limit {get;}
     PagingDirection Direction {get;}
}
```

What do you think?

clement911:
I guess this is not related directly to this issue and should be discussed on https://github.com/nozzlegear/ShopifySharp/issues/380

PR #394, opened 2019-09-13
Changed DiscountAllocation.DiscountApplicationIndex from int to long

clement911:
I have witnessed several orders from different stores where the discount_allocations[0].discount_application_index is a long (larger than int.MaxValue)
!
To be sure, this is a bug in Shopify because the discount_application_index is supposed to be the index of a valid application in discount_applications and there is not more than int.MaxValue values in this array.
Still, there is no way to deserialize these invalid orders unless the property is changed to a long.
Otherwise, ShopifySharp fails with the following error: 

`Newtonsoft.Json.JsonReaderException: JSON integer 66200961046 is too large or small for an Int32. Path 'line_items[0].discount_allocations[0].discount_application_index', line 1, position 4890.`

### DIFF
--- a/ShopifySharp/Entities/DiscountAllocation.cs
+++ b/ShopifySharp/Entities/DiscountAllocation.cs
@@ -14,7 +14,7 @@ namespace ShopifySharp
         /// The index of the associated discount application in the order's discount_applications list.
         /// </summary>
         [JsonProperty("discount_application_index")]
-        public int DiscountApplicationIndex { get; set; }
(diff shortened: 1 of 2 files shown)


nozzlegear:
Thanks! Sometimes I wish .NET just had a plain `number` type :sweat_smile:

nozzlegear:
Published in 4.21.5

clement911:
Thank you very much!

PR #396, opened 2019-09-18
Variant.grams can be larger than int.MaxValue so changing to long

clement911:
Otherwise the variant / order cannot be deserialized.

Also removed the NullToZeroConverter, which was introduced back when the grams property was non nullable.

Incrementing minor version


nozzlegear:
Published in 4.21.6!

clement911:
Thanks for the quick publish!

PR #432, opened 2019-12-31
Added read_locations enum in AuthorizationScope

ajayak:
https://help.shopify.com/en/api/getting-started/authentication/oauth/scopes


nozzlegear:
This has been published in version 4.25.0 on Nuget, which should be available within the next 30 minutes. Thank you very much!

PR #436, opened 2020-01-14
Shopify Refund API

PaulThomasRowe:
A service for creating Shopify Refunds.

**Please note**: For the associated test, OrderId is tied to a manually created, active order is required to be able to use this test. This is because it seems API refunds cannot be applied against orders created via the API.


nozzlegear:
Thank you! I'm going to go through these pull requests tomorrow and merge/publish them.

PR #437, opened 2020-01-15
Hmac bug

clement911:
Fixed HMAC calculation when query string includes array parameters, such as ids[] given when using a bulk action app link.

I included a test that I verified locally but that I marked as `Skip` because the HMAC depends on whatever secret key is configured on the build server.

@nozzlegear  if you can generate a real query string with the matching key then you can replace the hard coded query string and unskip the test. You'll need to create a bulk action link (https://help.shopify.com/en/api/embedded-apps/app-extensions/shopify-admin/bulk-action-links) with the app matching the secret key.

I hope this make sense?

### DIFF
--- a/ShopifySharp.Tests/Authorization_Tests.cs
+++ b/ShopifySharp.Tests/Authorization_Tests.cs
@@ -1,6 +1,7 @@
 using System;
 using System.Collections.Generic;
 using System.Threading.Tasks;
+using Microsoft.AspNetCore.WebUtilities;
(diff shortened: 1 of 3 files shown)


nozzlegear:
Thanks! That does make sense, I'll merge and publish these pull requests tomorrow and work on getting a testable querystring for this one.

nozzlegear:
Published in 4.25.0, thank you!

PR #441, opened 2020-01-18
Update restock in Refund.cs

Khungersumit:
SDK V4.25.0  Refund.cs File
Issue:  while creating refund exceptin:  {"error":"restock is no longer supported.
Please use restock_type on refund_line_items."}

FIX: Update restock_type in Refund.cs 
Reference: https://help.shopify.com/en/api/reference/orders/refund

Please review the changes :)

### DIFF
--- a/ShopifySharp/Entities/Refund.cs
+++ b/ShopifySharp/Entities/Refund.cs
@@ -59,12 +59,12 @@ namespace ShopifySharp
         /// </summary>
         [JsonProperty("refund_line_items")]
         public IEnumerable<RefundLineItem> RefundLineItems { get; set; }
-
+        
         /// <summary>
-        /// Whether or not the line items were added back to the store inventory.
+        /// How this refund affects inventory levels.
         /// </summary>
         [JsonProperty("restock_type")]
-        public bool? Restock { get; set; }
+        public string RestockType { get; set; }
 
         /// <summary>
         /// The list of <see cref="Transaction"/> objects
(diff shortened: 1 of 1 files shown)


ShopifySharp/Entities/Refund.cs
```diff
@@ -59,12 +59,12 @@ public class Refund : ShopifyObject
...
         [JsonProperty("refund_line_items")]
         public IEnumerable<RefundLineItem> RefundLineItems { get; set; }
-
+        
         /// <summary>
-        /// Whether or not the line items were added back to the store inventory.
+        /// How this refund affects inventory levels.
         /// </summary>
-        [JsonProperty("restock")]
-        public bool? Restock { get; set; }
```
nozzlegear: While this `Restock` property is no longer valid, let's leave it in and just decorate it with an obsolete attribute which will let everyone who's using it know they should use `RestockType` instead. Then I'll remove the `Restock` property in 5.0.

So something like this:

```cs
/// <summary>
/// Previously this property would indicate whether the line items were added or should be added back to the store. It is now obsolete and ignored by Shopify.
/// </summary>
[JsonProperty("restock")]
[Obsolete("The Restock property is now obsolete and will be removed in a future release. Attempting to use this property when creating or updating a Refund will result in the Shopify API returning an error.")]
public bool? Restock { get; set; }
```

Khungersumit:
r

PR #442, opened 2020-01-22
Line item property names can be arbitrary javascript objects.

clement911:
I wasn't aware of that until ShopifySharp failed to deserialize a recent order with an arbitrary javascript objects as the line item property name!
I tested and the API indeed accepts arbitrary object.

This is just changing the string to an object so that deserialization works.
In most case this will be a string and the user can just do `property.Name.ToString()`

image

image

@nozzlegear I would appreciate it if you're able to include this in the next release!


nozzlegear:
Published in 4.25.1!

clement911:
Thank you!

PR #443, opened 2020-01-24
Added [JsonIgnore] on Restock property because of refund error

ajayak:
Fixes Error
> Issue: while creating refund exceptin: {"error":"restock is no longer supported. Please use restock_type on refund_line_items."}

### DIFF
--- a/ShopifySharp/Entities/Refund.cs
+++ b/ShopifySharp/Entities/Refund.cs
@@ -10,7 +10,7 @@ namespace ShopifySharp
     public class Refund : ShopifyObject
     {
         /// <summary>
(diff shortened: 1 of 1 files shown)


ajayak:
Hi @nozzlegear, having the `restock` property causes errors even if we mark it obselete. Only way is `[JsonIgnore]` 😃

ajayak:
Hi @nozzlegear, please check this PR. The `RefundService` throws an error because of `restock` property in request. The only way to fix it is to remove this property from json.

nozzlegear:
Published in 4.25.2!

PR #456, opened 2020-02-24
v5 contrib

clement911:
Link header parsing logic:
-Extracted link header parsing logic into its own class
-Added tests for link header parsing
-Handling extra cases such as the presence of commas in the link url when the 'fields' parameter is present

### DIFF
--- a/ShopifySharp/Filters/IListFilter.cs
+++ b/ShopifySharp/Filters/IListFilter.cs
@@ -16,11 +16,6 @@ namespace ShopifySharp.Filters
         int? Limit { get; set; }
 
         /// <summary>
-        /// A comma-separated list of which fields to show in the results. This parameter only works for some endpoints.
-        /// </summary>
-        string Fields { get; set; }
-
-        /// <summary>
         /// Converts the filter to a list of key/value pairs which can be turned into a querystring.
         /// </summary>
         IEnumerable<KeyValuePair<string, object>> ToQueryParameters();
--- a/ShopifySharp/Filters/IListFilter.cs
+++ b/ShopifySharp/Filters/IListFilter.cs
@@ -1,3 +1,4 @@
+using Newtonsoft.Json;
 using System.Collections.Generic;
 
 namespace ShopifySharp.Filters
@@ -13,6 +14,11 @@ namespace ShopifySharp.Filters
         /// The number of items which should be returned. Default is 50, maximum is 250.
         /// </summary>
         int? Limit { get; set; }
+
+        /// <summary>
+        /// A comma-separated list of which fields to show in the results. This parameter only works for some endpoints.
+        /// </summary>
+        string Fields { get; set; }
 
         /// <summary>
         /// Converts the filter to a list of key/value pairs which can be turned into a querystring.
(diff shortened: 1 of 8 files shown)


clement911:
Using version 2019-10.
See #380 (comment) for why it is not using 2020-01

ShopifySharp/Filters/IListFilter.cs
```diff
@@ -14,6 +15,11 @@ public interface IListFilter
         /// </summary>
         int? Limit { get; set; }
 
+        /// <summary>
+        /// A comma-separated list of which fields to show in the results. This parameter only works for some endpoints.
+        /// </summary>
+        string Fields { get; set; }
```
nozzlegear: I think I want to leave `Fields` off of the base list filter, as not all of the endpoints using pagination make use of the `Fields` property. For v5 I'm going more with the philosophy of creating dedicated list filters for each service, and the dedicated ones add the properties (like fields, created_at_min, etc.) that the service supports. 

In v4 I went a little too overboard with DRY and tried to make every filter extend other filters as much as possible, even if the endpoints didn't support some of the properties. I want to get away from that in v5.

ShopifySharp/Lists/LinkHeaderParser.cs
```diff
@@ -0,0 +1,46 @@
...
+        private static Regex _regexPrevLink = new Regex(@"<(https://[^>]*)>\s*;\s*rel=""previous""", RegexOptions.Compiled | RegexOptions.CultureInvariant);
+        private static Regex _regexNextLink = new Regex(@"<(https://[^>]*)>\s*;\s*rel=""next""", RegexOptions.Compiled | RegexOptions.CultureInvariant);
+
+        public static LinkHeaderParseResult Parse(string linkHeaderValue)
+        {
+            var prevLink = GetPageInfoParam(linkHeaderValue, _regexPrevLink);
+            var nextLink = GetPageInfoParam(linkHeaderValue, _regexNextLink);
+
+            if (prevLink == null && nextLink == null)
+                throw new ShopifyException($"Found neither a 'previous' or 'next' url in the link header: '{linkHeaderValue}'");
```
nozzlegear: I'm not sure if throwing an exception is a good idea yet. I haven't tested this, but I could imagine a scenario where you request the first page of e.g. orders, and your shop only has 1 total order so there's no next page or previous page to link to.

ShopifySharp/Filters/IListFilter.cs
clement911: I think this new philosophy is the right way to go, however Shopify does clearly state that 3 parameters only can be changed when paginating: page_info, limit and fields.
https://shopify.dev/tutorials/make-paginated-requests-to-rest-admin-api

fields is unlikely to be changed from one page to the next though. I will remove it.

ShopifySharp/Lists/LinkHeaderParser.cs
clement911: I hesitated on this one however I left it in the end because Shopify stated that if there are no more pages, then the Link header will not be present at all.
So, if a Link header is present, it MUST have either a previous or next page.
Let me know if you still want to remove, but IMO the check is useful as it asserts that the parsing was succesful.

nozzlegear: Ah okay, I didn't know that! I think it's fine to leave it in then.

clement911: cool

ShopifySharp/Filters/IListFilter.cs
clement911: I removed the Fields property.

clement911:
Any more changes required?

nozzlegear:
None off the top of my head, but I'll be free later this afternoon to review and merge.

PR #457, opened 2020-02-24
PR to discuss v5.

clement911:
Hi @nozzlegear 

I am opening a draft pull request to discuss a few design choices related to v5.
This is solely for the purpose of discussing changes before actually making any real change.

I've opened a [separate PR](https://github.com/nozzlegear/ShopifySharp/pull/456) for changes that I think are ready to merge.

### DIFF
--- a/v5-discuss.txt
+++ b/v5-discuss.txt
@@ -1,2 +1,5 @@
 Do we really need IListResult<T>? It seems to be ListResult<T> is enough and doesn't require an interface
+
 Do we really need IListFilter? It seems to be ListFilter is enough and doesn't require an interface
+
+Why is ListFilter not inheriting from Parameterizable? Doing so would mean it could inherity from ToParameters() for free
--- a/v5-discuss.txt
+++ b/v5-discuss.txt
@@ -1,0 +1,1 @@
+Do we really need IListResult<T>? It seems to be ListResult<T> is enough and doesn't require an interface
--- a/v5-discuss.txt
+++ b/v5-discuss.txt
@@ -3,3 +3,5 @@ Do we really need IListResult<T>? It seems to be ListResult<T> is enough and doe
 Do we really need IListFilter? It seems to be ListFilter is enough and doesn't require an interface
 
 Why is ListFilter not inheriting from Parameterizable? Doing so would mean it could inherity from ToParameters() for free
+
+Why are Count methods commented out?
--- a/v5-discuss.txt
+++ b/v5-discuss.txt
@@ -1,1 +1,2 @@
 Do we really need IListResult<T>? It seems to be ListResult<T> is enough and doesn't require an interface
+Do we really need IListFilter? It seems to be ListFilter is enough and doesn't require an interface


nozzlegear:
Looks good so far! I added a couple comments on things I might want to change. I like the class you're using for next/previous page links, it's nice having both the `Url` and the `PageInfo` value.

clement911:
Good to hear. I have another potential PR once you accept the previous one.
Have you seen the questions I added to this PR (under Files changed I added a text file with some questions for you)

nozzlegear:
Woops, I didn't see those questions!

> Do we really need IListResult<T>? It seems to be ListResult<T> is enough and doesn't require an interface

This one is probably worthless since the service is always using `ListResult<T>` when returning, and there's nowhere the end user takes control of that. 

> Do we really need IListFilter? It seems to be ListFilter is enough and doesn't require an interface

I'd like to keep this one since end users are going to be the ones instantiating/passing these in to the services. I'm sure in almost all cases they'll just use the built-in ones, and they could easily override the `ListFilter<T>` if they wanted to make their own, so this is just being "csharpy" and using interfaces for the sake of it. 

> Why is ListFilter not inheriting from Parameterizable? Doing so would mean it could inherity from ToParameters() for free

My plan was to make each `ToQueryParameters` function "manually" (actually shared extension methods in most cases) build their own query parameter dictionaries instead of relying on reflection to do it. No particular reason for that, just to clean things up and remove `JsonProperty` as that attribute is semantically out of place on the list filters; they're never serialized to JSON. 

> Why are Count methods commented out?

Most are still intact, but I commented out some of them because they need their own count filter classes. A lot were sharing the same filters with their list methods, where some of the properties didn't apply to counting.

nozzlegear:
> My plan was to make each ToQueryParameters function "manually" (actually shared extension methods in most cases) build their own query parameter dictionaries instead of relying on reflection to do it. No particular reason for that, just to clean things up and remove JsonProperty as that attribute is semantically out of place on the list filters; they're never serialized to JSON.

Thinking about this more, I might make the base `ListFilter<T>` class use the `Parameterizable.ToParameters()` under the hood, and worry about removing those attributes after v5 is released.

clement911:
> Thinking about this more, I might make the base ListFilter<T> class use the Parameterizable.ToParameters() under the hood, and worry about removing those attributes after v5 is released.

I'd say this is wise given the time remaining. Already a lot has changed so I'd say it's best not to change too much else unless needed. We can always revise this in the next minor version.

nozzlegear:
This seems like a better place to discuss progress on v5 as a whole. I've just finished refactoring all of the services to use the new paginated list processes. I'm going to merge your new pull request soon, then get to work on fixing the count methods/adding count filters. After that I'll be able to publish a prerelease. 

I think the biggest blocker before I can publish v5 is going to be refactoring the tests to work with the new list filter/results classes. I don't want to assign you any work to do, but if you have some free time to refactor a few tests that would help me tremendously!

clement911:
Happy to help buy could you please fix the compilation errors and merge the other PR first?

nozzlegear:
Will do! Finishing up the count filters right now.

nozzlegear:
Merged your PR! I believe we're ready to do a prerelease.

PR #460, opened 2020-02-26
V5 contrib3

clement911:
Mostly cleanup.

There is still a lot of unnecessary/duplicated plumbing code. (I didn't have time to go through all services.)

What's the point of UnpaginatedListFilter? Non paginated filter don't have any common logic like that Paginated filters do...


nozzlegear:
I haven't checked yet, but the unpaginated list endpoints might still support a limit parameter, which I'd add to the unpaginated list filter. But even if it doesn't, I want all list methods to accept a filter that can be turned into query params. That will let devs use undocumented filter parameters if they discover them, such as the secret `name` parameter for listing orders.

clement911:
> I haven't checked yet, but the unpaginated list endpoints might still support a limit parameter, which I'd add to the unpaginated list filter.

Some, but not all, will accept a `limit` param. In line with your earlier comment about not trying to be too DRY, the `limit` param should be explicitely added only on concrete filters for which the service does support it.

> But even if it doesn't, I want all list methods to accept a filter that can be turned into query params. That will let devs use undocumented filter parameters if they discover them, such as the secret name parameter for listing orders.

Concrete unpaginated list filters could simply inherit from `Parameterizable` directly for that. And users can easily inherit from concrete filter classes as well.

clement911:
I removed IListFilter in a separate commit in case you insist on keeping it, you could simply revert the commit...

nozzlegear:
I don't think keeping it is necessary, it just adds overhead and almost everyone who wants to modify the filters will just inherit from the existing ones. I did see a couple small changes that I might want to make before merging this, I'll do a review thing in a few minutes.

clement911:
It's not going to be possible to 2 methods both named `ListAsync` in a given service because the compiler won't be able to resolve the correct overload. I would suggest renaming the paging ones to `PageAsync`.
So, we would have a `ListAsync` and a `PageAsync`. `ListAsync` can call `PageAsync`.

nozzlegear:
Oh sorry, I'm not sure I understand! Why would there be two ListAsync methods?

clement911:
For example in `ProductService`:
`public virtual async Task<ListResult<Product>> ListAsync(ListFilter<Product> filter)`
`public virtual async Task<ListResult<Product>> ListAsync(ProductListFilter filter)`

clement911:
Same for `CountAsync`

nozzlegear:
Ah I see, let me think about that. I'd prefer to keep `ListAsync` instead of introducing a `PageAsync`, and I also want users to be able to create a class with _just_ the properties they need (plus limit/pageinfo); that way they'd have direct control over exactly what is getting serialized, without needing to create their own service implementation. Although we could also achieve that if we make the `ToQueryParameters` virtual and overridable.

clement911:
One could inherit a class from `ListFilter<T>` for that, no?

nozzlegear:
Oh yeah. I think I must be confused still! So I'm fine with removing the _interfaces_ for these list filters, and want to keep the `ListFilter<T>` classes. 

So if we have the product service and it has these two methods:

```cs
public virtual async Task<ListResult<Product>> ListAsync(ListFilter<Product> filter) {...}

public virtual async Task<ListResult<Product>> ListAsync(ProductListFilter filter) {...}
```

That works, right? The compiler should be able to tell if you're trying to use `ListAsync(new ProductListFilter())` and `ListAsync(new ListFilter<Product>)` I think?

clement911:
Oh yes you're right, I think I must be the one confused.
If both filter arguments were optional, there would be an issue, but I suppose there is no reason to make the `ListFilter<T>` parameter optional.

nozzlegear:
Okay great! So if we remove the `IUnpaginatedListFilter`, we'll need to change those services that used it to accept some kind of filter even if there are no documented properties for the service. I don't mind using the Parameterizable class for that, but I'd rather just have a wrapper around it that makes it obvious the class will be used for filtering results.

clement911:
Fair enough. This wrapper should probably be usable for non list filters too.

nozzlegear:
Sure! That sounds good to me.

nozzlegear:
Let me know when you're ready to merge and I'll do a quick review. The CI is building this branch, so we'll be able to do a prerelease today as soon as we've reached a consensus on the method signatures and classes.

clement911:
I've been adding a couple of methods to the base service class because all List and Count method repeat the same pattern over and over again.
I gotta head but I hope you can take a look at this PR.
I'm down to 38 compile errors in tests (I couldn't finish all entities yet)
Hopefully it's not too huge!! There is no logic change. Mostly mechanical refactorings...

It's pretty laborious work but I think v5 will set a good foundation for the future. I do like how you create explicit filter classes for all methods.

If you feel comfortable, please do merge this PR so that when I resume work the next PR is not as big.

nozzlegear:
Will do, thank you!

ShopifySharp/Infrastructure/Parameterizable.cs
```diff
@@ -58,7 +55,7 @@ public abstract class Parameterizable
         /// <param name="value">The property's value.</param>
         /// <param name="property">The property itself.</param>
         /// <returns>The new parameter.</returns>
-        protected virtual KeyValuePair<string, object> ToSingleParameter(string propName, object value, PropertyInfo property)
+        private KeyValuePair<string, object> ToSingleParameter(string propName, object value, PropertyInfo property)
```
nozzlegear: Same thing here, this should probably be protected virtual so devs can take control of it where necessary.

ShopifySharp/Services/CustomerSavedSearch/CustomerSavedSearchService.cs
```diff
@@ -23,61 +23,29 @@ public class CustomerSavedSearchService : ShopifyService
...
-        {
-            return await _CountAsync(filter);
-        }
-
         /// <summary>
         /// Gets a count of all of the shop's customers.
         /// </summary>
         /// <returns>The count of all customers for the shop.</returns>
-        public virtual async Task<int> CountAsync(CustomerSavedSearchCountFilter filter = null)
+        public async Task<int> CountAsync(CustomerSavedSearchCountFilter filter = null)
```
nozzlegear: Should be virtual so it can be overridden.

ShopifySharp/Services/DraftOrder/DraftOrderService.cs
```diff
@@ -17,61 +17,29 @@ public class DraftOrderService : ShopifyService
...
-            var response = await ExecuteRequestAsync<int>(req, HttpMethod.Get, rootElement: "count");
-
-            return response.Result;
-        }
-
-        /// <summary>
-        /// Retrieves a count of the shop's draft orders. 
-        /// </summary>
-        /// <param name="filter">Options for filtering the count.</param>
-        public virtual async Task<int> CountAsync(ICountFilter filter = null)
```
nozzlegear: We probably still want a method that accepts a generic filter, just like the list methods accept a generic ListFilter.

ShopifySharp/Services/User/UserService.cs
```diff
@@ -24,15 +24,10 @@ public UserService(string myShopifyUrl, string shopAccessToken) : base(myShopify
         /// <summary>
         /// Gets all the users.
         /// </summary>
-        public virtual async Task<IEnumerable<User>> ListAsync(IUnpaginatedListFilter<User> filter = null)
+        public virtual async Task<IEnumerable<User>> ListAsync()
```
nozzlegear: As discussed, all ListAsync methods should allow passing in a filter so devs can use undocumented properties if they desire.

nozzlegear:
Now that I'm actually looking at the tests and usage of these methods, I'm having second thoughts about allowing filters where they aren't documented. It implies that there's something you can pass in there to filter the results, but then you instantiate a new `CountFilter`, see it has no properties on it, and wonder what's wrong. 

I think the best approach is what you touched on, just add/allow filters for the endpoints that we _know_ use them. Then, by making the `ToQueryParameters` method virtual on the filter classes, devs can override that method to get direct control of what is being sent if they need to. For methods that don't have any documented filters, they can extend the service class itself to add one.

nozzlegear:
With that in mind, does it make sense to remove the extra `ListAsync` methods that take a `ListFilter<T>`? I think it does but want to hear your opinion first since you've done a lot of work on that detail.

nozzlegear:
Maybe not, I see now how it works. Since you can't change any filter properties when getting next/previous pages, those list methods that accept `ListFilter<T>` are strictly for getting the next/previous pages, and the ones that accept a dedicated filter like `OrderListFilter` are for getting the first page. I think that's a good design so I won't mess with it.

clement911:
> I think the best approach is what you touched on, just add/allow filters for the endpoints that we know use them. Then, by making the ToQueryParameters method virtual on the filter classes, devs can override that method to get direct control of what is being sent if they need to. For methods that don't have any documented filters, they can extend the service class itself to add one.

As the user, wouldn't it just be easier to inherit from the filter class to add additional properties?
Any non set properties on the base filter wouldn't be serialized anyway..

nozzlegear:
In almost all cases, yes. It still might be useful in very rare cases for them to directly control the serialization though, in case they find some bug or strange situation.

PR #461, opened 2020-02-26
Line item property deserialization

clement911:
I hit another weird case where the line_item.properties was an empty object instead of an array 😲!

image

### DIFF
--- a/ShopifySharp.Tests/Serialization_Tests.cs
+++ b/ShopifySharp.Tests/Serialization_Tests.cs
@@ -1,0 +1,56 @@
+using Newtonsoft.Json;
+using ShopifySharp.Infrastructure;
+using ShopifySharp.Lists;
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using System.Net;
+using System.Threading.Tasks;
+using Xunit;
+using EmptyAssert = ShopifySharp.Tests.Extensions.EmptyExtensions;
+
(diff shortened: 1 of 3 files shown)


nozzlegear:
image

How bizarre!

nozzlegear:
Just thinking out loud, what if we turn the object into an array? 

So this:

```json
{ "foo": "bar", "baz": "bat" }
```

Becomes this:

```cs
[
   new LineItemProperty { Name = "foo", Value = "bar" },
   new LineItemProperty { Name = "baz", Value = "bat" }
]
```

clement911:
Could do but the converter will get quite complicated.
And what if the object contains nested object (at this point I assume anything is possible!).

nozzlegear:
I think if there are nested objects, we'd just turn them into a dictionary or something. `LineItemProperty.Value` is `object` right? So we'd just read the top layer of keys/values and turn everything else into objects. 

That does sound like a lot of work, so I'm happy to merge this in right now and then take a crack at doing this converter myself later on unless you want to give it a shot before I merge.

clement911:
Right, I think that would work.
Would be great if you could merge and we can implement your idea in the near future.

nozzlegear:
Sure! One quick change first, can you add a remarks thing on the property with a link to this pull request?

```cs
/// <remarks>https://github.com/nozzlegear/ShopifySharp/pull/461</remarks>
```

clement911:
Done.

Issue #462, opened 2020-02-27
Version 5.0 release information


nozzlegear:
:tada: ShopifySharp v5.0 has been released and published on nuget! You can install it with the dotnet CLI like this:

```sh
dotnet add package ShopifySharp --version 5.0.3
```

I've [written a guide here for migrating from v4 to v5](https://nozzlegear.com/shopify/shopifysharp-version-5-migration-guide). It doesn't cover every little change, but it does cover the big ones that are most likely to cause you issues when upgrading. If you think something is missing from this guide, let me know down below. 

[**ShopifySharp version 5 migration guide.**](https://nozzlegear.com/shopify/shopifysharp-version-5-migration-guide)

You can use this thread to ask about changes, report bugs, and let me know if a class or property you were previously using is missing.

# Paginated listing

The big change in this version is the new paginated listing operations that Shopify is now requiring. You can no longer request page 1, page 2, page 3, etc. of orders/customers/whatevers. You must use a special parameter they return during list operations to get the next page or the previous page. If you do not use that parameter, you will only ever get the first page of results. 

Since listing all customers/orders is a common question I get, here's how you do that going forward in ShopifySharp v5.0:

```cs
var allOrders = new List<Order>();
var service = new OrderService(domain, accessToken);
var list = await service.ListAsync(new OrderListFilter 
{ 
    Limit = 250 
});

while (true)
{
    allOrders.AddRange(list.Items);

    if (!list.HasNextPage)
    {
        break;
    }

    list = await service.ListAsync(list.GetNextPageFilter());
}
```

A couple important things to note about the above example:

- You must use the list result's `LinkHeader` to get a filter that will let you request the next page or the previous page. 
- The link header may be null, so you must check it before using it. This can happen e.g. if you're on page one and there's only one page total.
- The next page link and previous page link headers may specifically be null, even if the link header itself is not. This can happen e.g. when you're at the end of the list, there will be no next page link, but there will be a previous page link, and vice versa.
- **Once you've listed the first page, you cannot change any filter properties except for `Limit` and `Fields`.** Beyond the first page, Shopify will "remember" the filter properties you used previously and continue to apply them to all subsequent pages. If you want to make a change to one of those properties, you need to create a new filter and start over with the first page. 

# ~TODO~ Implemented

(**This section was tracking changes that needed to be made before the v5.0 release. They are now obsolete as v5.0 has been published.**)

Things to consider/implement before the full release is published:

[...]

PR #463, opened 2020-02-27
V5 contrib 4

clement911:
Fixed all tests compile errors + migrated more services to use the Service base class methods.

### DIFF
--- a/ShopifySharp.Tests/CustomerSavedSearch_Tests.cs
+++ b/ShopifySharp.Tests/CustomerSavedSearch_Tests.cs
@@ -93,28 +93,6 @@ namespace ShopifySharp.Tests
         }
 
         [Fact]
-        public async Task Searches_For_CustomerSavedSearch()
-        {
-            // It takes anywhere between 3 seconds to 30 seconds for Shopify to index new CustomerSavedSearch for searches.
-            // Rather than putting a 20 second Thread.Sleep in the test, we'll just assume it's successful if the
-            // test doesn't throw an exception.
-            bool threw = false;
-
-            try
-            {
-                var search = await Fixture.Service.SearchAsync("-notes");
-            }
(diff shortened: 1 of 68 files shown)


clement911:
https://github.com/nozzlegear/ShopifySharp/issues/462

clement911:
Alright, I was able to test cursor pagination and it appears to work nicely.
All the tests now compile and I think you can merge and release a new pre-release version.
Are you able to run all the tests and report which one fail?

ps:
I added new base service methods to DRY out POST/PUT/DELETE.
So far only migrated Webhook and RecurringCharge services.
It should be trivial to migrate other services.

clement911:
At this point I implemented almost everything I had in mind mind.

Regarding 'Fields', I still think we should add it to `ListFilter<T>`. This way it can be auto loaded automatically like 'Limit' does. What do you think?

Also, what about renaming `ListFilter<T>` to `PageFilter<T>`?

clement911:
I converted many service methods to use the base service methods (including POST/DELETE/PUT), but not all of them.
If you have some energy left, feel free to have a go :) Otherwise, this might not be critical and can be migrated over time...

clement911:
Ready for merge.

nozzlegear:
> At this point I implemented almost everything I had in mind mind.
> 
> Regarding 'Fields', I still think we should add it to `ListFilter<T>`. This way it can be auto loaded automatically like 'Limit' does. What do you think?
> 
> Also, what about renaming `ListFilter<T>` to `PageFilter<T>`?

Yes I think we should add the `Fields` property to the filter so it can be autoloaded. I'm working on the package right now and will try to fit this in. As for renaming the class, I've thought about it and I like `ListFilter` more than `PageFilter`, simply because we're using it in `ListAsync` methods so it fits the name a little better.

PR #465, opened 2020-03-02
Contrib next

clement911:
- `MetaFieldFilter.UpdatedAtMax` should be nullable.
- Added tests to ensure all Filter classes properties are nullable.


clement911:
We could potentially add a similar test to ensure all entity properties are nullable...

nozzlegear:
That's a clever test! Good catch on the recurring charge URL too.

PR #466, opened 2020-03-02
Retry only on 429 caused by leaky bucket

clement911:


nozzlegear:
It looks like this breaks the fulfillment tests, since they're one of the things with the dev shop rate limit. Any suggestion on fixing them? I'm thinking just using the "dumb" retry policy maybe.

clement911:
The dumb retry policy was also enhanced to not retry on non leaky bucket breaches.
How about applying a retry or wait in the tests directly?

clement911:
Or maybe we could add a constructor bool parameter to the the policies, to specific whether it should retry on non bucket breach?

nozzlegear:
I like the constructor approach, that way we can default to this new behavior but make it possible to use the old behavior in case somebody was relying on it. Would you be able to create a pull request for that?

clement911:
Will do shortly.

clement911:
see https://github.com/nozzlegear/ShopifySharp/pull/467

nozzlegear:
Published in 5.0.1, thank you!

PR #467, opened 2020-03-02
Add retry policy ctor param to whether to only retry when leaky b uck…

clement911:
…et is full.

I haven't had a chance to test but code is quite simple.


nozzlegear:
Awesome, thank you! I'll update the tests to set the boolean to false.

PR #470, opened 2020-03-04
Fixing Fields property encoding

StriveDen:
Fix to fields encoding issue mentioned in:
[https://github.com/nozzlegear/ShopifySharp/issues/380#issuecomment-594217876](https://github.com/nozzlegear/ShopifySharp/issues/380#issuecomment-594217876)

Fields query parameter returned from Shopify has "," character changed to "%2C". As a result of this all object properties are null when using next or previous links.

### DIFF
--- a/ShopifySharp/Lists/LinkHeaderParser.cs
+++ b/ShopifySharp/Lists/LinkHeaderParser.cs
@@ -38,11 +38,12 @@ namespace ShopifySharp.Lists
                 throw new ShopifyException($"Cannot parse page link url: '{matchedUrl}'");
             }
 
-            string GetQueryParam(string name)
-            {
-                return uri.Query.Split('?', '&')
-                                .FirstOrDefault(p => p.StartsWith($"{name}="))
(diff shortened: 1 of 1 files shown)


clement911:
Rather than hard coding for `fields` and `,` a better way would be to url-decode the value before returning in `GetQueryParam`.

This way, it would work for all parameters (e.g. there is a comma in the `page_info`) and for all characters (e.g. there is `<` or `>`)

clement911:
Looks better!
Although you formatted your code strangely for the `?.` operator

StriveDen:
> Looks better!
> Although you formatted your code strangely for the `?.` operator

Oh, you right. I've installed some weird code formatter in VS Code that made this. Should be ok now.

nozzlegear:
Good catch! I'm going to merge this now, but I want to write a test for it this afternoon before publishing to nuget. I'll update here once it's published.

nozzlegear:
Test passes, this has been published in 5.0.4. Thank you for the pull request!

PR #476, opened 2020-03-09
Introduce Cancellation Token support across all public service APIs.  (Resolves #238)

### LINKED ISSUE #238: Provide CancellationToken as optional parameter to all methods, so requests can be cancelled or timed-out.
It is common for APIs involving potentially long http requests, to accept a `CancellationToken`, such that a call can be cancelled, or timed-out by a user.
It would be fairly simple to add to this API, as it would be just need be passed down to the `HttpClient`, that accepts one in `SendAsync`.

pianomanjh:
- `ShopifyService` ExecuteXXX calls leverage a cancellation token for the underlying `HttpClient` request and the execution policy timeout
- All existing Service public API methods now take an optional `CancellationToken`

I included some tests on the ShopifyService just to ensure the token is leveraged properly.  I did not add tests to every single service API call, as they proxy requests to the base class, and token use there is tested.  Therefore, please review the public service calls to ensure I didn't miss anything.

### DIFF
--- a/ShopifySharp.Experimental/ApplicationCredit.fs
+++ b/ShopifySharp.Experimental/ApplicationCredit.fs
@@ -1,6 +1,7 @@
 ﻿namespace ShopifySharp.Experimental
 
 open System.Net.Http
+open System.Threading
 open ShopifySharp
 open ShopifySharp.Infrastructure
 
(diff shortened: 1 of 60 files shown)


clement911:
Nice!

nozzlegear:
Awesome, thank you for all the work on this! I know it was a lot of tedious changes. Sorry it took so long to merge in, I'll let the tests run and get it published soon.

clement911:
thank you @pianomanjh  for your contribution!

nozzlegear:
Just a heads up, this seems to be randomly throwing randomly throwing both `OperationCanceledException` and `TaskCanceledException`, it's not consistent at the moment. Not sure if that's an implementation issue, or if it's just the nature of cancelling tasks.

pianomanjh:
Yeah, that is likely a difference between the exceptions thrown by consumers of the token.  We pass it into both a `Task.Delay` and `HttpClient`.  Could vary by whichever one 'notices' the cancellation first.  I'll take a better look at it tonight and report back.

nozzlegear:
Published in 5.1.0 on nuget. Thanks for the PR and all the work you put into this one!

Issue #475, opened 2020-03-06
Product UpdateAsync - could not update variants to []


madushann:
We are getting and exception when try to update product without any variants. Is that the expected behavior of the end point? If so how can we remove the variants.

`could not update variants to []   at ShopifySharp.ShopifyService.CheckResponseExceptions(HttpResponseMessage response, String rawResponse)
   at ShopifySharp.ShopifyService.<>c__DisplayClass21_0`1.<<ExecuteRequestAsync>b__0>d.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at ShopifySharp.DefaultRequestExecutionPolicy.Run[T](CloneableRequestMessage request, ExecuteRequestAsync`1 executeRequestAsync)
   at ShopifySharp.ShopifyService.ExecuteRequestAsync[T](RequestUri uri, HttpMethod method, HttpContent content, String rootElement)
   at ShopifySharp.ProductService.UpdateAsync(Int64 productId, Product product)`

nozzlegear:
As far as I know this is not expected, but it’s possible Shopify has changed the behavior in the API version we’re now targeting. Can you post the code you’re using to update the product?

madushann:
Below is how we call the service end point. 

`ProductService productService = new ProductService(channel.Host, secondary ? channel.SecondaryToken : channel.Token);

shopifyProduct = await productService.UpdateAsync((long)shopifyProduct.Id, shopifyProduct);`

nozzlegear:
Where is that `shopifyProduct` coming from, and which properties are you trying to change when updating it? That information will help me track down where the problem may lie.

madushann:
It has all the information such as, 
Title, Options, BodyHtml, ProductType, Variants, etc. variants property is empty as below,
![image](https://user-images.githubusercontent.com/40527502/76136371-515f1200-5ff6-11ea-9820-aabef4e4f178.png)

nozzlegear:
I see, if you set the variants property to null before updating, does that change anything?

madushann:
@nozzlegear yes it worked when you set the variants property to null, when its empty it gives the error.

simplenotezy:
I have the same issue. We cannot set variants property to null, as we need to update the product variants.

Any updates on why this occour?

Issue #388, opened 2019-08-06
Experimental: explicitly building object properties for create/update methods


nozzlegear:
I'm working on a ShopifySharp.Experimental package that I plan to publish to Nuget in the next few days. I want to use this package as a testing ground for new ideas that won't necessarily make it into ShopifySharp on the first iteration. I won't promise stability in this package between releases, I expect there will be breaking changes quite often, so don't use it in a production app. 

The first experiment I'm doing is to fix the major issue this package has with updating/creating objects (#379) (#373) (#367) (#284). You have to be able to send null to the Shopify API, but it's far too easy to send null for a property that you didn't intend. In practice, I've take to a hard rule that first I need to pull in the entire Shopify object, set the one single property I want to change on it, and then send that entire object back to Shopify -- just to make sure I don't accidentally erase a field. 

In my experiment, I'm setting up an F# module which will explicitly build each property of an object so you know _exactly_ what will be sent when creating and updating. Null will never be sent if you don't explicitly add the property and set it to null. Other properties will not be sent at all if you don't add them. 

Here's what the intended usage looks like once I've got this finished:

```fs
open ShopifySharp.Experimental 

let webhook = 
	newWebhook
	|> address "https://example.com"
	|> topic "app/uninstalled"

// Serialized to JSON, this looks like:
// { "address": "https://example.com", "topic": "app/uninstalled" }
```

Note that the `format`, `metafield_namespaces`, `created_at`, `updated_at` fields are not null like they would be if you serialized the regular ShopifySharp.Webhook class -- they just aren't in the JSON at all. 

If you want to set a property to null, you'll use `makePropertyNull`:

```fs
let webhook = 
	newWebhook
	|> address "https://example.com"
	|> topic "app/uninstalled"
	|> makePropertyNull WebhookProperty.Format

// Serialized to JSON, this looks like:
// { "address": "https://example.com", "topic": "app/uninstalled", "format": null }
```

If you need to remove a property from the object for whatever reason, you'll use `removeProperty`:

```fs
let webhook = 
	newWebhook
	|> address "https://example.com"
	|> topic "app/uninstalled"
	|> format "json"
	|> removeProperty WebhookProperty.Format

// Serialized to JSON, this looks like:
// { "address": "https://example.com", "topic": "app/uninstalled" }
```

[...]

Issue #379, opened 2019-07-03
Partial Product Update Broken


lantram:
Using a sample store with a product catalog from here ("jewelry.csv"):

https://github.com/shopifypartners/product-csvs

If I do this:

```
await productService.UpdateAsync(product.Id.Value,
    new Product()
    {
        Title = "New product title"
    });
```

It messes up the store. In particular, it makes the product unavailable (removes the online store sales channel). I don't know if anything else is affected.

Instead I have to do this:

```
// get product from productService
product.Title = "New product title";
await productService.UpdateAsync(product.Id.Value,product);
```

The documentation on **Updating a product** at https://github.com/nozzlegear/ShopifySharp#products says we can do this:

```
var service = new ProductService(myShopifyUrl, shopAccessToken);
var product = await service.UpdateAsync(productId, new Product()
{
    Title = "New product title"
});
```

However, that is not currently correct since it appears to be broken.

nozzlegear:
I'm currently working on a ShopifySharp.Experimental package, with the main goal of introducing a new way to create/update objects without accidentally sending values you didn't intend to send (including null). I'll update here when it's published.

shafaqat-ali-cms365:
I am having the same issue, I had 5000 products and I updated all of them and now all products have been removed form the the Online Store channel. Please help, all products have been unlinked from the online store channel.

Any temporary fix or anything, please help.

nozzlegear:
Hey @shafaqat-ali-cms365, this issue is partially out of date as it doesn't contain the fix for the problem. I'm keeping it open because I'm still planning a better, more encompassing fix in 6.0. 

The problem is, because of the nature of C# and Json.Net, fields that you don't explicitly set on an object will get serialized to their default C# value. For a nullable date field, that value is `null` which unpublishes your products. When updating the product, you need to explicitly set the `PublishedAt` value so it doesn't get serialized to null:

```cs
var product = await service.UpdateAsync(new Product
{
    PublishedAt = someDateTimeValue,
    // And other updated values here 
});
```

PR #477, opened 2020-03-09
Update Product by explicit fields

drovani:
I know there's work being done in F# (#388) for specifying which fields get sent back to in an Update call. In the meantime, I've quickly put together a way to pick which fields to use by listing them in a string array. The idea being that one doesn't need to do a GET to pull down the whole Product before doing an Update with just one or two fields.

For immediate application, I am using the Product Service. I have added a couple of test methods to validate that it is working. In an ideal setup, I might allow for an expression tree, but implementing that is outside of my current knowledge set.

Any feedback you have would be appreciated. I wouldn't mind expanding this to more services as needed.

This would solve #379 and possibly #475.


nozzlegear:
This is cool! I especially like your use of `nameof`, that wasn't something I would have considered but now that I see it it's obvious.

I am indeed doing some work in the same area with F#, but I've recently started a new branch for bringing a fluent-style builder to C# and will instead base the F# stuff off of that. My intent is for it to look something like this:

```cs
var product = new ProductBuilder()
	.SetTitle("Some title")
	.SetPublishedAt(DateTimeOffset.Now);
```

It would do essentially what your pull request is doing too, it just modifies a hidden dictionary and decides which properties get serialized. That said, I definitely like your approach too, and I think having the two different approaches is beneficial, so I'd like to go ahead with your proposal. 

Two thoughts after looking at the code:

1. I'd like to explore using both a blacklist and a whitelist to explicitly decide which props get serialized. They'd be exclusive, so if both a whitelist and a blacklist are used, prefer the whitelist. 
2. Eventually the parameters and overloads for each method on each service are going to get out of hand, especially when I merge the PR with cancellation token support (#476). Maybe the create and update methods should accept a wrapper class that contains both the object itself, and another property to specify which fields should be serialized.

clement911:
Personally I prefer this approach to have a `SetXXX` method for each property (`SetTitle`, `SetPublishedAt`).

-  Instead of putting this logic in the `ProductService`, it would be best to centralize it in `ShopifyService` by adding an additional optional `fields` parameter to the `ExecutePutAsync` method: https://github.com/nozzlegear/ShopifySharp/blob/e1a09d1544de3188ddd00caeb8ae15a636aaae64/ShopifySharp/Services/ShopifyService.cs#L235

All service update methods can then simply call this method.

- Instead of a simple `string[] fields`, we could create a `FieldsInfo` class.
Pseudo code:

```
enum FieldsMode { GivenFieldsOnly, AllExceptGivenFields }
class FieldsInfo
{
   string[] Fields;
   FieldsMode Mode;
}
```

This way we can ask to update all fields except a few.

- I also think it's worth exploring an Expression Tree version.
I don't think it would be to hard to convert an expression to a list of string, or maybe even a list of PropertyInfo, which is what ToDictionary does with the fields array.

`ExecutePutAsync` could accept a lamda instead of string[]

drovani:
> * Instead of putting this logic in the `ProductService`, it would be best to centralize it in `ShopifyService` by adding an additional optional `fields` parameter to the `ExecutePutAsync` method: https://github.com/nozzlegear/ShopifySharp/blob/e1a09d1544de3188ddd00caeb8ae15a636aaae64/ShopifySharp/Services/ShopifyService.cs#L235
> 

This brings up an interesting question - why is the `WebhookService` the only service that calls `ExecutePutAsync<T>`? The `ProductService` calls `ExecuteRequestAsync<T>`, which seems to duplicate work that `ExecuteWithContentCoreAsync<T>` already does.

https://github.com/nozzlegear/ShopifySharp/blob/e1a09d1544de3188ddd00caeb8ae15a636aaae64/ShopifySharp/Services/Product/ProductService.cs#L98

clement911:
I recently introduced base methods in the ShopifyService.cs, such as `ExecutePutAsync` and `ExecuteGetAsync`, because all services basically copy and paste the same logic.

We were quite pressed to release v5 and I didn't have time to go through all services but ideally they should all use the base class methods.

drovani:
I've put together this idea of a generic `UpdateBuilder<TShopifyObject>` that allows for a fluid set of property expressions and values. It stores a dictionary internally to keep track as the user builds the data points to update. I think this is cleaner than making a `Set_______` method for each property (I don't even want to think of the amount of boilerplate code that would require), and keeps things a little tighter than relying on the user to ensure that the property strings match.

I then added an overload for the `ProductService.UpdateAsync` to accept a `UpdateBuilder<Product>`, which then gets passed to the `ShopifyService.ExecutePutAsync`.

I'm not particularly happy with the duplication of the code that gets the `JsonPropertyAttribute`, since it's just a copy/paste from part of the `ObjectExtensions` class. Any suggestions on where that code could live?

I haven't tested this yet for nested properties (i.e. Update Product.Variants[].Ids), that'll be an iteration if there's positive feedback on this `UpdateBuilder<T>` object.

clement911:
I saw that Stripe.net has come up with [an interesting approach](https://github.com/stripe/stripe-dotnet/pull/1956) using a discriminated union `AnyOf<T1,T2>` and a special `Empty` parameter that can be used to explicitly set a field to null.. If we added that to all fields that can be set to null, and change the serialization settings to understand `AnyOf<T1,T2>`, this might work. Something to think about....

drovani:
It's interesting, but would require a fundamental rewrite of all of the ShopifySharp POCO classes, wouldn't it?

Is there interest in me further expanding on this `UpdateBuilder<TShopifyObject>` class, or do you think there's a better way to do this without significant change required on the part of consumers? I have some free cycles this week, and wouldn't mind spending it plugging away on this project.

PR #486, opened 2020-03-20
Graph leaky bucket

andre719mv:
Hi! 
I implemented Leaky Bucket for GraphQL. I am very limited in time so I did not formatted well my PR. Also I am pretty sure that you will not use my exact implementation. But if you will want to implement Leaky Bucket you may find something usefull here.

### DIFF
--- a/ShopifySharp/Services/ShopifyService.cs
+++ b/ShopifySharp/Services/ShopifyService.cs
@@ -14,35 +14,39 @@ namespace ShopifySharp
 {
     public abstract class ShopifyService
     {
-        public virtual string APIVersion => "2019-10";
+        public virtual string APIVersion => "2020-01";
+        protected virtual bool SupportsAPIVersioning => true;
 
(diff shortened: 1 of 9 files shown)


nozzlegear:
Hey, thanks for the PR! Unfortunately I don't think I can merge it as is, because there's a lot of changes that go beyond just adding a leaky bucket to the graph service. For example, there's a lot of changes to the ShopifyService.cs file which comment out code and add some debug prints instead.

ShopifySharp/Services/ShopifyService.cs
```diff
@@ -14,35 +14,39 @@ namespace ShopifySharp
 {
```
nozzlegear, line 14: This file currently has too many unrelated changes for me to merge, particularly the debug logger printing.

andre719mv:
Hey!
Sure. As I said before this PR is only to see the idea. I did not have time to prepare something that can be merged.

nozzlegear:
Cool, thanks! I'll close the PR for now but will come back to look at it when we add leaky bucket support for the graph API.

PR #502, opened 2020-04-25
Fix a incorrect jsonProperty position

yinghualuowu:
It should be in this position
image


nozzlegear:
Published in 5.1.1. Thank you!

PR #512, opened 2020-05-12
added support for multipass url generator

knopa:
I am not sure how better to do the test. Technically best option is to find the user in Shopify customers by multipass identifier to assert it.


nozzlegear:
Thanks for the pull request @knopa! I don't think I'm familiar with this feature, can you explain what this is for, and add a link to some documentation?

knopa:
It's from list unimplemented functionality 
https://help.shopify.com/api/reference/multipass

knopa:
It's needed when you need to allow auto-create customers from your existing project not Shopify one

ShopifySharp/Services/Multipass/MultipassService.cs
```diff
@@ -0,0 +1,99 @@
+﻿using Newtonsoft.Json;
+using System;
+using System.IO;
+using System.Linq;
+using System.Security.Cryptography;
+using System.Text;
+
+namespace ShopifySharp
+{
+    public class MultipassService
```
nozzlegear, line 10: It looks like the only method on this class is `GetMultipassUrl`, would it make more sense to make the class static, remove the constructor, and then make the method take `myShopifyUrl` and `multipassSecret` in addition to `customer`?

nozzlegear (changes requested):
Hey! I'm looking into finally merging this into master. I've left a review comment about changing the class to static, would you mind taking a look if you're still interested in getting this merged?

ShopifySharp/Services/Multipass/MultipassService.cs
knopa: Hey, I have updated as your asked.

knopa:
Hey, I have updated as your asked.

nozzlegear (approved):

nozzlegear:
Released in 5.18.0 on nuget! Thanks for your contribution!

PR #517, opened 2020-05-25
Changed InventoryLevel.Available from int to long

clement911:
I came across in production an inventory_level with a available quantity lower than int.MinValue

image


nozzlegear:
Thanks for the PR! I'm not sure I'm comfortable with altering the exception message right now. It's probably a bad practice, but it's likely that some people are checking the message value when catching exceptions, as I used to recommend they do that for rate exceptions. I think it should still be possible to access that request ID if you try to cast a generic exception to a ShopifyException?

clement911:
Fair enough.
I've reverted that part.

nozzlegear:
Thanks! Published in 5.2.0

clement911:
Thank you

PR #531, opened 2020-06-30
Fix for IsAuthenticRequest bug when ids[] contains a single value.

### LINKED ISSUE #530: IsAuthenticRequest is not working when the query string contains array parameters and the value has only one element.
This is related to #437 

When the query string contains array parameters like ids[] and the value for those parameter has only one value then the IsAuthenticRequest fails to generate the correct HMAC. 

Sample query

GET /xxxx/xxxxxx?hmac=b63470b593885527be43c4ebdbfa2d66d9862ea69208d292ea4cafaeb7911357&ids[]=2544739451040&locale=en&session=cf0f2004fb006d2ec33810114105416a5d34172ad831185e31e8b136a86232f1&shop=xxxxx.myshopify.com&timestamp=1593490902
#
I have modified the EncodeQuery method to fix this issue

private static string PrepareQuerystring(IEnumerable<KeyValuePair<string, StringValues>> querystring, string joinWith)
{
	var kvps = querystring.Select(kvp => new
	{
		Key = EncodeQuery(kvp.Key, kvp.Key, true),
		Value = EncodeQuery(kvp.Key, kvp.Value, false)
	})
		.Where(kvp => kvp.Key != "signature" && kvp.Key != "hmac")
		.OrderBy(kvp => kvp.Key, StringComparer.Ordinal)
		.Select(kvp => $"{kvp.Key}={kvp.Value}");

	return string.Join(joinWith, kvps);
}

private static string EncodeQuery(string key, StringValues values, bool isKey)
{
	//array parameters are handled differently: see https://community.shopify.com/c/Shopify-APIs-SDKs/HMAC-calculation-vs-ids-arrays/td-p/261154

	string s = "";
	if (!isKey && key.Contains("[]"))
	{
		s = '[' + string.Join(", ", values.Select(v => '"' + v + '"')) + ']';
	}
	else
		s = values.Count <= 1 ? values.FirstOrDefault() : '[' + string.Join(", ", values.Select(v => '"' + v + '"')) + ']';

	if (string.IsNullOrEmpty(s))
	{
		return "";
	}

	//Important: Replace % before replacing &. Else second replace will replace those %25s.
	string output = (s.Replace("%", "%25").Replace("&", "%26")) ?? "";

	if (isKey)
	{
		output = output.Replace("=", "%3D").Replace("[]", "");
	}

	return output;
}

clement911:
Fixes #530

@nozzlegear  if possible, complete the tests with a real query string value. I cannot do it myself as I don't know the keys used by the test server.

### DIFF
--- a/ShopifySharp.Tests/Authorization_Tests.cs
+++ b/ShopifySharp.Tests/Authorization_Tests.cs
@@ -75,7 +75,16 @@ namespace ShopifySharp.Tests
         [Fact(Skip = "TODO: Generate a real query string with the shop and secret key used by the build server, which contains an ids[] parameter")]
         public void Validates_Web_Requests_WithArrayParameter()
         {
-            string query = "hmac=123....";
+            string query = "hmac=...";
+            var qs = QueryHelpers.ParseQuery(query);
+            bool isValid = AuthorizationService.IsAuthenticRequest(qs, Utils.SecretKey);
+            Assert.True(isValid);
+        }
+
+        [Fact(Skip = "TODO: Generate a real query string with the shop and secret key used by the build server, which contains an ids[] parameter with a single value")]
+        public void Validates_Web_Requests_WithArrayParameter_SingleValue()
+        {
+            string query = "hmac=...";
             var qs = QueryHelpers.ParseQuery(query);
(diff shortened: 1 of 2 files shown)


nozzlegear:
Published in 5.4.0!

PR #533, opened 2020-07-02
Bug Fix for GiftCard.ExpiresOn

ChaoticIke:
When ~~creating~~ updating a gift card with an expiry date, the value is ignored because the date format needs to be 'yyyy-MM-dd'

The Date Format Converter takes a format parameter since it comes from one of my own projects. Let me know if you want to change it.

### DIFF
--- a/ShopifySharp.Tests/GiftCard_Tests.cs
+++ b/ShopifySharp.Tests/GiftCard_Tests.cs
@@ -84,14 +84,21 @@ namespace ShopifySharp.Tests
         }
 
         [Fact(Skip = "Cannot run without a Shopify Plus account.")]
-        public async Task Creates_GiftCards_With_ExpiresOn()
+        public async Task Updates_GiftCards_With_ExpiresOn()
         {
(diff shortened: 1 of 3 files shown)


ChaoticIke:
image
I had some trouble getting to run the tests using the Environment Variables. I just hard coded my Shopify credentials for the tests.

It seems that the Create API doesn't accept an ExpiresOn, so I have changed the test to update instead.

I also needed to change the search test since searching by code isn't supported in this version of the API

nozzlegear:
Sorry, I thought I had merged this and published it but it seems I did not!

nozzlegear:
Alright, published in 5.4.0 on Nuget. Thank you for the pull request!

clement911:
@ChaoticIke Shouldn't the ExpiresOn be a DateTime instead of a DateTimeOffset then? Since it doesn't have any time information...

ChaoticIke:
@clement911 I thought about that as well. I wasn't sure if you would want to introduce a breaking change to fix a serialisation issue, so I left the data types as they were.

PR #541, opened 2020-07-20
Fix compile time error

PtrStruct:
It's going to throw you a compile time error if you try assigning it to a string. I suspect that this hasn't been updated in a while where as the actual library has. I'm also letting the user know how to generate their own API key by referencing the official documentation.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -260,8 +260,11 @@ var scopes = new List<string>()
     "write_customers"
 }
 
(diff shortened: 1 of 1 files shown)


nozzlegear:
Ah thanks! I've been meaning to get the documentation updated, some of it is out of date after 5.0. I keep pushing it off because I want to move it all to it's own site, I need to get that done.

PR #548, opened 2020-07-23
Updated the readme with an example

PtrStruct:
Updated the readme with an example of how to create a product with a variant in one go, and setting the variant title.


nozzlegear:
Awesome, thanks for the PR!

PR #567, opened 2020-08-25
List Fulfillment orders

pylenius:
We need way to list orderlines which should be fulfilled at specific location.

### DIFF
--- a/ShopifySharp/Entities/FulfillmentOrder.cs
+++ b/ShopifySharp/Entities/FulfillmentOrder.cs
@@ -1,0 +1,78 @@
+using System.Collections.Generic;
+using Newtonsoft.Json;
+
+namespace ShopifySharp
+{
+    /// <summary>
+    /// An object representing a Shopify fulfillment order.
+    /// </summary>
+    public class FulfillmentOrder : ShopifyObject
+    {
+        /// <summary>
+        /// The unique numeric identifier for the shop.
+        /// </summary>
+        [JsonProperty("shop_id")]
+        public long? ShopId { get; set; }
+
+        /// <summary>
+        /// The unique numeric identifier for the order.
+        /// </summary>
+        [JsonProperty("order_id")]
+        public long? OrderId { get; set; }
+
+        /// <summary>
+        /// The unique numeric identifier for the location.
+        /// </summary>
+        [JsonProperty("assigned_location_id")]
+        public long? LocationId { get; set; }
+
+        /// <summary>
+        /// "mars-fulfillment".
+        /// </summary>
+        [JsonProperty("fulfillment_service_handle")]
+        public string FulfillmentServiceHandle { get; set; }
+
+        /// <summary>
+        /// "submitted".
+        /// </summary>
+        [JsonProperty("request_status")]
+        public string RequestStatus { get; set; }
+
+        /// <summary>
+        /// "open".
+        /// </summary>
+        [JsonProperty("status")]
+        public string Status { get; set; }
+
+        /// <summary>
+        /// Destination for the fulfillment.
+        /// </summary>
+        [JsonProperty("destination")]
+        public Address Destination { get; set; }
+
+        /// <summary>
+        /// A historical record of each item in the fulfillment.
+        /// </summary>
+        [JsonProperty("destination")]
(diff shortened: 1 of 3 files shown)


ShopifySharp/Entities/FulfillmentOrder.cs
```diff
@@ -0,0 +1,78 @@
...
+        /// <summary>
+        /// The unique numeric identifier for the order.
+        /// </summary>
+        [JsonProperty("order_id")]
+        public long? OrderId { get; set; }
+
+        /// <summary>
+        /// The unique numeric identifier for the location.
+        /// </summary>
+        [JsonProperty("assigned_location_id")]
```
clement911, line 26: Name of property should match name in API => AssignedLocationId

ShopifySharp/Entities/FulfillmentOrder.cs
```diff
@@ -0,0 +1,78 @@
...
+        /// "open".
+        /// </summary>
+        [JsonProperty("status")]
+        public string Status { get; set; }
+
+        /// <summary>
+        /// Destination for the fulfillment.
+        /// </summary>
+        [JsonProperty("destination")]
+        public Address Destination { get; set; }
```
clement911, line 51: Wondering if it might be better to create a new class `FulfillmentOrderDestination`.
Even if the `Address` class matches all fields, they might drift apart in the future...

ShopifySharp/Entities/FulfillmentOrder.cs
```diff
@@ -0,0 +1,78 @@
...
+
+        /// <summary>
+        /// A historical record of each item in the fulfillment.
+        /// </summary>
+        [JsonProperty("line_items")]
+        public IEnumerable<LineItem> LineItems { get; set; }
```
clement911, line 57: We need a dedicated class, such as `FulfillmentOrderLineItem`

ShopifySharp/Entities/FulfillmentOrder.cs
```diff
@@ -0,0 +1,78 @@
...
+        /// <summary>
+        /// "create_fulfillment", "request_fulfillment", "cancel_fulfillment_order", "request_cancellation"
+        /// </summary>
+        [JsonProperty("supported_actions")]
+        public IEnumerable<string> SupportedActions { get; set; }
+
+        /// <summary>
+        /// ???
+        /// </summary>
+        [JsonProperty("merchant_requests")]
```
clement911, line 68: Need a `FulfillmentOrderMerchantRequest` class

ShopifySharp/Services/FulfillmentOrders/FulfillmentService.cs
```diff
@@ -0,0 +1,40 @@
...
+        /// <param name="myShopifyUrl">The shop's *.myshopify.com URL.</param>
+        /// <param name="shopAccessToken">An API access token for the shop.</param>
+        public FulfillmentOrderService(string myShopifyUrl, string shopAccessToken) : base(myShopifyUrl, shopAccessToken) { }
+
+        /// <summary>
+        /// Gets a list of up to 250 of the order's fulfillments.
+        /// </summary>
+        /// <param name="orderId">The order id to which the fulfillments belong.</param>
+        /// <param name="cancellationToken">Cancellation Token</param>
+        public virtual async Task<ListResult<FulfillmentOrder>> ListAsync(long orderId, CancellationToken cancellationToken = default)
```
clement911, line 34: The results of this endpoint are not paginated. From memory, ListResult must only be used for paginated results.

clement911:
Thanks for the PR, I've added a few comments.
Also, `shop_id` property is missing and we need some unit tests ideally.

Finally, I think it would be better to merge this in the 2020-07 branch (which will become master next).
@nozzlegear  what do you think? Again, I really don't multiple having multiple endpoints with different versions. Very hard to follow and version classes...

YorkshireWill:
Hi All,

Are there plans to merge this to base branch? Would save me some time if this was in the main library rather than current work I'm doing to extend this for FulfillmentOrders (I'm not comfortable enough with GIT yet - still a TFS user at the moment or I'd help contribute to this)

Thanks,
Will

nozzlegear:
Closed in favor of #588

PR #568, opened 2020-09-03
2020-07 contrib

clement911:
The library is now pinned to 2020-07.
I've added the new fields introduced in 2020-04.
I believe this is ready for merge into master, since @nozzlegear already implemented the new `CollectionService` 👍.

I would consider incrementing the major version, just because the `CollectService.ListAsyc` method will now fail when passing the id of a smart collection.


clement911:
@nozzlegear  are you able to merge?

nozzlegear:
Yep, sorry! Planning to publish tomorrow. I need to get back on my open source work schedule!

PR #592, opened 2020-10-27
Added additional Order properties

jvoce:
Added Properties to the Order and ShippingLine entities to return additional presentment money data


nozzlegear:
Published on nuget in ShopifySharp version 5.8.0 -- thank you for contributing!

PR #595, opened 2020-11-05
add "source" to transaction.cs

juanm21:
source: The origin of the transaction. Set to external to create a cash transaction for the associated order.


nozzlegear:
Published on nuget in version 5.8.0, thank you for contributing!

PR #608, opened 2021-01-20
Added primary location id in Shop object.

mattetom:


nozzlegear:
Published in version 5.8.1 on Nuget. Thank you!

PR #624, opened 2021-03-11
Maximum refundable

### LINKED ISSUE #553: "maximum_refundable" missing from Refund.Shipping and Refund.Transactions
maximum_refundable is a property returned from the refunds/calculate api.
[Api Reference](https://shopify.dev/docs/admin-api/rest/reference/orders/refund?api[version]=2019-10#calculate-2019-10).

At the moment the Refund object uses the same Transaction object as the other APIs so I wasn't sure if it would be appropriate to add the property there or if a new inheriting object would be better.

Any thoughts?

Edit:
Examples
[Adding property directly](https://github.com/nozzlegear/ShopifySharp/compare/master...ChaoticIke:MaximumRefund)
[Adding via inheritance](https://github.com/nozzlegear/ShopifySharp/compare/master...ChaoticIke:MaximumRefundNewObject)
[Inheritance for only the calculate method](https://github.com/nozzlegear/ShopifySharp/compare/master...ChaoticIke:MaximumRefundCalculateOnly)

### LINKED ISSUE #555: RefundService.Calculate returns 400 when not specifying the optional options parameter
Usage:
`refundService.CalculateAsync(orderId)`

Error (Status 400):
```json
{
    "errors": {
        "refund": "Required parameter missing or invalid"
    }
}
```

I was thinking this could be made into a required parameter, the null check could be changed or the refund class could use an include null attribute. The following payloads returned a response:

```json
{
  "refund": {
    "refund_line_items": null
  }
}

{
  "refund": {
    "refund_line_items": []
  }
}

{
  "refund": {
    "shipping": null
  }
}
```

Partially related:
#553 
In this state the response would return the shipping and transactions sections with 0 amount
```json
{
    "refund": {
        "shipping": {
            "amount": "0.00",
            "tax": "0.00",
            "maximum_refundable": "9.99"
        },
        "refund_line_items": [],
        "transactions": [
            {
                "order_id": --,
                "kind": "suggested_refund",
                "gateway": "bogus",
                "parent_id": --,
                "amount": "0.00",
                "currency": "GBP",
                "maximum_refundable": "44.98"
            }
        ],
        "currency": "GBP"
    }
}
```

ChaoticIke:
- Adds maximum_refundable to Shipping and Transaction
- Make refund options required rather than trying to generate a default
- Adds discrepancy_reason to Refund
Fixes #553
Fixes #555

### DIFF
--- a/ShopifySharp/Entities/Refund.cs
+++ b/ShopifySharp/Entities/Refund.cs
@@ -52,6 +52,14 @@ namespace ShopifySharp
         public string Note { get; set; }
 
         /// <summary>
+        /// An optional comment that explains a discrepancy between calculated and actual refund amounts. 
+        /// Used to populate the reason property of the resulting order adjustment object attached to the refund.
+        /// </summary>
+        /// <value>restock, damage, customer, and other.</value>
+        [JsonProperty("discrepancy_reason")]
+        public string DiscrepancyReason { get; set; }
+
+        /// <summary>
         /// The list of <see cref="RefundLineItem"/> objects
         /// </summary>
         [JsonProperty("refund_line_items")]
--- a/ShopifySharp/Entities/Refund.cs
+++ b/ShopifySharp/Entities/Refund.cs
@@ -96,5 +96,11 @@ namespace ShopifySharp
         /// </summary>
         [JsonProperty("amount")]
         public decimal? Amount { get; set; }
+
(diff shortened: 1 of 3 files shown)


nozzlegear:
Published in version 5.9.0 on Nuget. Thank you for the pull request!

PR #627, opened 2021-03-25
removed useCustomerDefaultAddress parameter from DraftOrder  Create m…

Uffman:
removed useCustomerDefaultAddress parameter from DraftOrder Create method

### DIFF
--- a/ShopifySharp/Services/DraftOrder/DraftOrderService.cs
+++ b/ShopifySharp/Services/DraftOrder/DraftOrderService.cs
@@ -7,7 +7,7 @@ using ShopifySharp.Filters;
 using ShopifySharp.Infrastructure;
 using ShopifySharp.Lists;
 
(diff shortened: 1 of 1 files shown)


nozzlegear:
Hey @Uffman, thank you for the pull request! I appreciate you contributing, but this change is a "breaking" change that can break the builds of people using the library. That's something that I usually only do in a major new release, e.g. ShopifySharp v6.0. If you want to create a new pull request and just add this change as an overload of the current method, I'd be happy to merge it:

```cs
public virtual async Task<DraftOrder> CreateAsync(DraftOrder order, bool useCustomerDefaultAddress = false, CancellationToken cancellationToken = default)
{
    // ...
}

public virtual async Task<DraftOrder> CreateAsync(DraftOrder order, CancellationToken cancellationToken = default)
{
    // ...
}
```

PR #629, opened 2021-03-30
Typos + 1 new draft order property

clement911:
Mostly typos + 1 new draft order property.

### DIFF
--- a/ShopifySharp/Entities/DraftOrder.cs
+++ b/ShopifySharp/Entities/DraftOrder.cs
@@ -97,6 +97,12 @@ namespace ShopifySharp
         public bool? TaxExempt { get; set; }
 
         /// <summary>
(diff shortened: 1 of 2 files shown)


clement911:
Hi @nozzlegear .
Are you able to merge and release?
Without date properties being datetimeoffset , they are automatically converted to the machine's timezone, which is probably not what is expected.
I don't believe it is a breaking change because DateTimeOffset implicitely converts to DateTime so it should not produce any compilation errors to update.

I also convert Price.Amount from a string to a decimal, which is a broken change, but that would have definitely been a typo that was introduced earlier.

nozzlegear:
Sure! I can merge it this morning.

--
Josh Harms

clement911:
Awesome!

clement911:
Thanks for the quick turn around!

PR #669, opened 2021-09-24
Various fixes and improvements for LeakyBucketPolicy

clement911:
Running the LeakyBucketPolicy in production revealed a few issues that I investigated and fixed.
Also, the GraphQL logic is now smarter as it handles 'refunds' (when the actual cost was less than the expected cost).


clement911:
Thanks!

nozzlegear:
Published in 5.13.1 on Nuget! Thanks for the fixes!

clement911:
Thanks!

PR #672, opened 2021-10-01
LeakyBucketExecutionPolicy support for requests context 

clement911:
This is a non breaking change. 
A new optional parameter `getRequestContext` can be supplied when constructing a `LeakyBucketExecutionPolicy`.

It serves to indicate whether the current context is `Foreground` or `Background`. The policy will prioritize foreground requests and run them first.

For example, if a user is waiting on the request (loading a page showing response results for example), you would want to prioritize this request over competing requests that might be running in the background (to run some jobs for example).
(In this scenario, you could for example inspect if there is an ambient HttpContext to determine whether the request is Foreground or Background)


nozzlegear:
Published in v5.14.0 on nuget!

Issue #685, opened 2021-11-15
LeakyBucketExecutionPolicy fails to handle graph error


nozzlegear:
Adding this here so I don't forget to add a unit test for this. The `LeakyBucketExecutionPolicy` class throws a null reference exception when it tries to parse this response from Shopify:

```json
{"errors":[{"message":"Field 'shopOwner' doesn't exist on type 'Shop'","locations":[{"line":8,"column":21}],"path":["query","shop","shopOwner"],"extensions":{"code":"undefinedField","typeName":"Shop","fieldName":"shopOwner"}}]}
```

Stack trace:

```
System.NullReferenceException: Object reference not set to an instance of an object.
   at ShopifySharp.LeakyBucketExecutionPolicy.Run[T](CloneableRequestMessage baseRequest, ExecuteRequestAsync`1 executeRequestAsync, CancellationToken cancellationToken, Nullable`1 graphqlQueryCost)
   at ShopifySharp.ShopifyService.ExecuteRequestAsync(RequestUri uri, HttpMethod method, CancellationToken cancellationToken, HttpContent content, Dictionary`2 headers, Nullable`1 graphqlQueryCost)
   at ShopifySharp.GraphService.SendAsync(RequestUri req, HttpContent content, Nullable`1 graphqlQueryCost, CancellationToken cancellationToken)
   at ShopifySharp.GraphService.PostAsync(String body, Nullable`1 graphqlQueryCost, CancellationToken cancellationToken)
   at FSharp.Control.Tasks.TaskBuilder.bindTaskConfigureFalse@131.Invoke(Unit unitVar0)
   at FSharp.Control.Tasks.TaskBuilder.StepStateMachine`1.nextAwaitable()
--- End of stack trace from previous location ---
   at FSharp.Control.Tasks.TaskBuilder.bindTaskConfigureFalse@131.Invoke(Unit unitVar0)
   at FSharp.Control.Tasks.TaskBuilder.StepStateMachine`1.nextAwaitable()
--- End of stack trace from previous location ---
   at FSharp.Control.Tasks.TaskBuilder.bindTaskConfigureFalse@131.Invoke(Unit unitVar0)
   at FSharp.Control.Tasks.TaskBuilder.StepStateMachine`1.nextAwaitable()
--- End of stack trace from previous location ---
   at Microsoft.Extensions.Hosting.Internal.Host.StartAsync(CancellationToken cancellationToken)
[...]

PR #686, opened 2021-11-15
Fix null ref exception in leaky bucket policy 

clement911:
Fixes https://github.com/nozzlegear/ShopifySharp/issues/685


clement911:
Big release! Thanks @nozzlegear

PR #690, opened 2021-11-18
Fix for serialization bug

clement911:
I started seeing errors in the logs as soon as I deployed the latest version.

```Newtonsoft.Json.JsonReaderException: Could not convert string to DateTimeOffset: 0000-12-31T18:09:24-05:50. Path 'accepts_marketing_updated_at'```

It turns out Shopify can populate an invalid date such as 0000-12-31T18:09:24-05:50, which cannot be parsed to a DateTimeOffset.

Luckily we already had a converter to handle this case so the fix was pretty simple.

### DIFF
--- a/ShopifySharp/Entities/Customer.cs
+++ b/ShopifySharp/Entities/Customer.cs
@@ -1,4 +1,5 @@
 using Newtonsoft.Json;
+using ShopifySharp.Converters;
 using System;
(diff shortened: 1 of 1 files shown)


nozzlegear:
Good catch! I'm surprised the tests didn't pick it up. Your fix is now published in v5.14.1 on nuget!

PR #706, opened 2022-01-03
New property line.TipPaymentGatewaySpecified to determine whether a l…

clement911:
…ine is a tip line.

### DIFF
--- a/ShopifySharp/Entities/LineItem.cs
+++ b/ShopifySharp/Entities/LineItem.cs
@@ -128,6 +128,14 @@ namespace ShopifySharp
         public string TipPaymentMethod { get; set; }
 
         /// <summary>
+        /// Whether the tip_payment_gateway field is present or not
+        /// If true, the line is a tip line
+        /// For a tip line, tip_payment_gateway is always specified (though it can be null)
+        /// For a non tip line, tip_payment_gateway is never specified
+        /// </summary>
+        public bool TipPaymentGatewaySpecified { get; set; }
+
+        /// <summary>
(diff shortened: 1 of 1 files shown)


clement911:
Thank @nozzlegear. Could you publish a new package?

clement911:
@nozzlegear ?

clement911:
Awesome. Thanks!

nozzlegear:
Hey @clement911, I can't seem to find this property in the docs. Should it be nullable, and does it need a `JsonProperty` attribute?

nozzlegear:
I've made it nullable and added the `JsonProperty` attribute for now. Let me know if that's wrong!

clement911:
@nozzlegear no this is not an API property and shouldn't be annotated with JsonProperty.
It's a built-in feature of json.net. The bool property will be set to true if present in the payload, otherwise false. It is useful to differentiate between a null value (true) and the absence of value (false), since the underlying property will be null in the .net object in both cases.

See https://stackoverflow.com/a/61532883

https://dotnetfiddle.net/Iyry36

https://www.newtonsoft.com/json/help/html/P_Newtonsoft_Json_Serialization_DefaultContractResolver_IgnoreIsSpecifiedMembers.htm

nozzlegear:
Oh neat! I don't think I've seen that before. I'll revert my changes.

nozzlegear:
Reverted in 5.14.4!

clement911:
Thanks 😀

clement911:
I noticed you added a JsonIgnore attribute. I'm not sure if this will have an impact.

clement911:
I tested and it works fine so we're all good 👍

PR #712, opened 2022-02-04
New class CustomerMarketingConsentWebhook

clement911:
A new webhook topic `customers_marketing_consent/update` was recently introduced.
This represents the payload data that is included in the webhook request, so that it can be deserialized to an object.


nozzlegear:
Published in 5.15.0!

clement911:
Thanks!

PR #724, opened 2022-03-04
Adding missing authorization scopes as defined in https://shopify.dev…

ajdvoynos:
Just a small change to add 2 scopes that were missing in the enum, and are defined at https://shopify.dev/api/usage/access-scopes


nozzlegear:
Released in 5.16.0 on nuget. Thanks for your patience!

PR #742, opened 2022-05-24
Added OrderEditWebhook entity

clement911:
Added OrderEditWebhook entity.
Unlike other orders/* topics, the orders/edited webhook doesn't contain a normal order payload.
Instead it contains info about the edit and what was changed. This entity is useful to deserialize the payload into an object.


nozzlegear:
Released in 5.16.0!

PR #745, opened 2022-05-31
Add Github actions build/test/publish/release pipeline

nozzlegear:
Replacing AppVeyor with Github actions workflow. 

TODO:

- [x] Pack projects separately from test projects.
- [x] Publish prerelease packages to ~~Github nuget feed~~ nuget.
  - [x] Is it possible to continue publishing these prerelease packages to Nuget without requiring approval each time?
  - [x] ~~The build numbers will become out of sync (i.e. Github build number starts at 1, but Appveyor is on 500+). Does that matter?~~ [Answer](https://github.com/nozzlegear/ShopifySharp/pull/745#issuecomment-1163799816)
- [x] Run test suite automatically on each push to master. 
  - [x] ~~Should this also run automatically for each pull request? Does it run on each commit? Is there a risk of exposing the access token? Can it run with some kind of command comment like `!test`?~~ [Answer](https://github.com/nozzlegear/ShopifySharp/pull/745#issuecomment-1145286727)
- [x] Improve test execution time (Github actions have a limit of 2000 per month and average unit test duration for ShopifySharp is 11-12 minutes). Run two or three categories at once? All at once and let the leaky bucket policy figure it out?
- [x] Run .NET Framework-specific tests (**not all tests**) on a Windows image. 
  - These tests currently fail on Ubuntu 20.04. I could probably get them working with a little digging, but it's just as easy to add a separate Windows step that only runs the .NET Framework-specific tests. 
  - [x] Properly categorize the .NET Framework tests so they can be run individually.
- [x] Parse out test category name from file itself instead of from the filename. 
  - The `AccessScope_Tests.cs` file is never tested because its category is `Access scopes` but the test script uses `AccessScope` from the filename as the category.
  - [x] Take into account that some test files will have more than one category (i.e. the DotNetFramework category tests).
- [x] For prereleases, create or append to a Github prerelease with links to all relevant issues/merged PRs.
  - [x] ~~What would the prerelease be named?~~ [Answer](https://github.com/nozzlegear/ShopifySharp/pull/745#issuecomment-1164445206)


nozzlegear:
> The build numbers will become out of sync (i.e. Github build number starts at 1, but Appveyor is on 500+). Does that matter?

This won't matter because the build numbers are only suffixes of whatever version is being published. 

Appveyor published `ShopifySharp v1.2.3-b500`, but Github Actions published `ShopifySharp v1.2.4-b1`. These won't ever overlap.

nozzlegear:
> What would the prerelease be named?

Prerelease will be named `Development build` with a tag of `latest`.

nozzlegear:
Manual workflow triggers can only be run when the yaml file is in the master branch. This pull request has completed all work that doesn't deal with full releases, so I'm merging this and splitting off the release-specific work into another pull request.

PR #749, opened 2022-06-15
Added ability to override the API version used when instantiating a GraphService

clement911:
ShopifySharp is pinned to an API version, which makes sense when using the REST API because the requets and responses objects are defined in the library itself.

However, for GraphService, it makes sense to be able to override the desired API version since the response is dictated by the shape of the request.
For example, right now I'd like to create a grapqhQL request with the latest stable version but I can't because ShopifySharp is pinned to an earlier version. (I know that I can override the GraphService and override the api version but I think that being able to override it in the constructor makes it a lot easier).


nozzlegear:
This has been released in 5.16.0 on nuget! Sorry about the wait!

PR #760, opened 2022-07-12
Added pre_tax_price and pre_tax_price_set properties

stephan-ansems:
Both the pre_tax_price and pre_tax_price_set properties are available on most Shopify Plus shops, but this is not (publicly) documented. By adding this it becomes possible to test and use the value without the need to calculate pre_tax_price values for products. Very useful in customs applications...


nozzlegear:
This has been published in version 5.16.0 on nuget. Thanks for the pull request!

Issue #764, opened 2022-07-19
The OrderTransactionId and UserId should be a long type in the GiftCardAdjustment


brian-winfield:
Both of these fields are ints in the GiftCardAdjustment class, which is now throwing exceptions because Shopify is now up to larger values for those fields. 

For example, I'm getting a cast exception:

`Could not convert to integer: 5756037562530. Path 'adjustments[0].order_transaction_id'.
`

The pull request for this fix is https://github.com/nozzlegear/ShopifySharp/pull/765

nozzlegear:
Fixed in 5.16.0 on nuget. Thanks!

PR #765, opened 2022-07-19
Change GiftCardAdjustment to use long instead of int. 

brian-winfield:
Changed the OrderTransactionId and the UserId to longs from int
Added the long values in the test


brian-winfield:
I also created an issue for this https://github.com/nozzlegear/ShopifySharp/issues/764

nozzlegear:
Released in 5.16.0 on nuget. Thanks for contributing!

Issue #776, opened 2022-08-28
Missing properties on FulfillmentOrder


davidkdb:
Hi,

These properties are missing on FullfillmentOrder:

fulfill_by
fulfillment_holds

nozzlegear:
Added in 5.18.3 on nuget!

PR #779, opened 2022-08-31
Fulfillment Order Missing Properties

judah4:
Fixes for #776.

Added fulfill_by and fulfillment_holds. Also fixed International Duties to be an object with incoterm.


nozzlegear:
Published in 5.18.3 on nuget. Thanks for the pull request!

PR #801, opened 2022-10-20
Update to API version 2022-07

nozzlegear:


clement911:
😁

nozzlegear:
For some reason I thought I had already done this! I was getting ready to do 2022-10 and realized I never got this one in like I thought.

clement911:
2022-10 might be a tricky one because of customer protection data new requirements.
Some Shopify users might not want to migrate until they get approved.

nozzlegear:
Ah that’s a good point, I’m not sure what the best way to handle that will be yet. My own app is going to be subject to the new protection requirements so I’ll need to get it ready, but if I remember correctly they gave me until July 2023 at the latest to upgrade. I think new apps have less time to do it.

--
Josh Harms

davidkdb:
DO NOT UPGRADE TO 2022-10!!

This will cause major issues for apps because of the customer protection data. It might take a while before all your users get this approval, and they have until July 2023 to get this.

Probably best to wait until March/April 2023

clement911:
We could potentially do the following:
-For 2022-10, create a new branch with an incremented major version
-For 2022-07, keep working on master, only incrementing minor version

This way, users could decide if they want to keep using 2022-07 for some time, until they are ready to move to 2022-10.

alexhauser:
It would indeed be great to get `2022-07` merged in time before `2022-04`'s deprecation on April 1st, 2023.

nozzlegear:
Thanks for the reminder! I'm working on merging and publishing this today.

nozzlegear:
All the tests pass for this, I'm just not quite sure where I left off so I don't know if this has all of the changes for 2022-07 incorporated yet.

clement911:
I had a look at https://shopify.dev/api/release-notes/2022-07 and I believe this PR already removes all deprecated endpoints.
So I believe we can merge this and release 2022-07. Any reason not to?

nozzlegear:
Thanks for checking! I agree, I had a look and it looks like everything deprecated is removed. And the new fulfillment endpoints are all implemented, tests are passing for them. Not sure why I never merged this.

clement911:
look good to me

clement911:
can we merge & release?

nozzlegear:
Not quite, got a few tests that started failing once I changed the API version to 2022-07. I'm fixing them right now, shouldn't be long!

--
Josh Harms

nozzlegear:
Okay, current status is all tests are passing except for the `ProductListing_Tests` and the `CollectionListing_Tests`. I won't be able to fix them tonight and I'm not sure if I'll have much time tomorrow. Any ideas on what might be broken here?

image

image

clement911:
I imagine it's because the app you're testing with is not configured an a sales channel.

image


It could be that they changed the behaviour in 2022-07 to return 404 instead of empty results?

PR #806, opened 2022-11-11
Fix for Charge.test serialization

clement911:
Ensure that `Charge.Test` and `RecurringCharge.Test` property is always serialized so that the `FalseToNullConverter` can kick in.

The `FalseToNullConverter` wasn't kicking in because the serializer uses `NullValueHandling.Ignore`
This is fixed by applying `NullValueHandling.Include` at the property level.

I personally think this converter should be removed altogether, but some people might rely on the fact that this the `Test` property should always return either false or true and never null. The API seems to accept false values.

Maybe we can remove `FalseToNullConverter.cs` altogether in the major/breaking release.

### DIFF
--- a/ShopifySharp.Tests/FalseToNullConverter_Tests.cs
+++ b/ShopifySharp.Tests/FalseToNullConverter_Tests.cs
@@ -1,5 +1,7 @@
 using Newtonsoft.Json;
+using Newtonsoft.Json.Linq;
 using ShopifySharp.Converters;
+using ShopifySharp.Infrastructure;
 using Xunit;
 
 namespace ShopifySharp.Tests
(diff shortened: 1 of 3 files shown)


nozzlegear:
I agree, this seems like a good one to remove in 6.0.

clement911:
thanks. looks like you did a great job with respect to the build pipeline and release workflow. very easy to use!

nozzlegear:
Glad you like it! I want to take another look at the pipeline soon so that both of us can approve full releases to nuget for Minor and Patch releases, but require reviews/approval for Major releases. I'm pretty sure that's possible but I need to review the github actions docs and figure out how to incorporate it in the scripts.

PR #808, opened 2022-11-15
Avoid using JObject.Parse because it doesn't preserve timezones

clement911:
Avoid using JObject.Parse because it doesn't preserve timezone on datetimeoffset (it always converts to the local timezone) See https://stackoverflow.com/questions/46829888/how-parse-string-to-jobject-ignoring-time-zone


clement911:
@nozzlegear  are you able to approve this one?

nozzlegear:
Done! Good catch on that bug, that doesn’t look like it was easy to track down.

--
Josh Harms

clement911:
@nozzlegear thank you very much.
That bug definitely caught me by surprise!

PR #826, opened 2023-01-10
Add ListForFulfillmentAsync method to FulfillmentService

thomotron:
This implements the ['Retrieves fulfillments associated with a fulfillment order' endpoint](https://shopify.dev/api/admin-rest/latest/resources/fulfillment#get-fulfillment-orders-fulfillment-order-id-fulfillments) from the REST API. It allows listing fulfilments from a particular fulfilment order.

This endpoint is available for for API versions 2022-04 and above.

The method arguments are ambiguous with `ListAsync`, so to distinguish the two this is named similarly to `CreateForFulfillmentAsync`.

The test requires a fulfilment order ID, so the FulfillmentService instance was roped in to deal with that.

I haven't added an entry to the README for this one but I'm happy to if needed.

### DIFF
--- a/ShopifySharp.Tests/Fulfillment_Tests.cs
+++ b/ShopifySharp.Tests/Fulfillment_Tests.cs
@@ -62,6 +62,17 @@ namespace ShopifySharp.Tests
         }
 
         [Fact]
+        public async Task Lists_Fulfillments_For_A_FulfillmentOrder()
+        {
+            long orderId = Fixture.Created.First().OrderId.Value;
+            var fulfillmentOrder = await Fixture.GetFulfillmentOrder(orderId);
+            long fulfillmentOrderId = fulfillmentOrder.Id.Value;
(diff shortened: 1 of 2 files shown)


nozzlegear:
Thanks @thomotron! Do you think this method would make more sense in the `FulfillmentOrderService` class, or where you have it now in the `FulfillmentService`? I think if I understand it correctly it makes more sense where you have it now, but maybe it should be renamed something like `ListForFulfillmentOrderAsync`?

thomotron:
Given it's returning fulfilments I would imagine it would be best under `FulfillmentService`. That's where it was listed under the API docs too so it supported the case in my head.

And yeah `ListForFulfillmentOrderAsync` sounds a lot better. I think I was a bit poorly timed with the latest release coming out a day or so after putting this up! I'll switch it over and push it through once I'm off the road for a sec.

nozzlegear (approved):

PR #841, opened 2023-02-17
PriceRule not respecting BOGO type rules + new gate scopes

zxed:
PriceRule was missing some additional id lists and attributes; resulting in BOGO type discount price rules being incorrectly created as 100% off discounts


zxed:
@nozzlegear planning to merge this into the code base?

CrimsonOrion:
Any word on this? I could really use it as well.

zxed:
P.s I have a fork, with updated branch merge. Should be public. You can
build that. That’s what I did for our app which needed to be on the more
recent api level.

nozzlegear:
Oh sorry, I meant to merge this with my last batch of merges. Will publish soon!

PR #859, opened 2023-03-23
Check env.yml for prefixed variables

alexcarson440:
Update Utils.Get() so that when checking for environment variables in the env.yml file that variables prefixed with 'SHOPIFYSHARP_' can be found.


nozzlegear (approved):

nozzlegear:
Good catch, thanks!

Issue #852, opened 2023-03-10
Filtering for location_ids on the AssigendFulfillmentOrder resource is not working


plogistik:
Hi,

The request URL ShopifySharp is building for the [AssignedFulfillmentOrder resource](https://shopify.dev/docs/api/admin-rest/2022-04/resources/assignedfulfillmentorder#resource-object) isn't correctly created if you add [location_ids as filter criteria](https://shopify.dev/docs/api/admin-rest/2022-04/resources/assignedfulfillmentorder#get-assigned-fulfillment-orders?assignment-status=cancellation-requested&location-ids[]=24826418). 

It is supposed to be `&location_ids[]=ANY_LOCATION_IDS seperated with commas`. The URL which gets generated builds the paramter as follows: &location_ids=ANY_LOCATION_IDS seperated with commas`. Due to the missing `[]` in the URL the request gets a 400 status code response from Shopify.

Following code I've used to built the filter object:
```
var assignedFulfullmentOrderServiceFilter = new AssignedFulfillmentOrderFilter
{
    AssignmentStatus = "fulfillment_requested",
    LocationIds = new long[]
    {
        interfaceConfig.InventoryLocationId.Value
    }
};
```

Has anyone encountered the same issue?

PR #860, opened 2023-03-24
Added square brackets to the location_ids json property name for AssignedFulfillmentOrderFilter

grantmorgandls:
As per the problem noted in Issue #852 

When calling to AssignedFulfillmentOrderService.ListAsync method with an AssignedFulfillmentOrderFilter containing a specific locationId, the query string needs square brackets to denote the array of location_ids.

Before this change, when attempting to filter AssignedFulfillmentOrders by location ID, a 400 status code response would be received with the following message.
`{
    "errors": {
        "location_ids": "expected String to be a Array"
    }
}`

Adding the square brackets to the end of the JsonPropertyName fixes this and it now filters by locationId correctly.


clement911:
Thank you!

Issue #853, opened 2023-03-13
SmartRetryExecutionPolicy removed without noticeable replacement/explaination


Skyhighjinks-Dev:
Basically - I am upgrading an application that used ShopifySharp 5.8.1 (or something old like that) and upgrading it to 6.0.1. I've noticed quite a few changes however I am most concerned about the `SmartRetryExecutionPolicy` removal in 6.0.0+

https://github.com/nozzlegear/ShopifySharp/commit/fda0977d58270ee4268701aa5b0164f9faa993e8

I am wondering if `ShopifyService.SetGlobalExecutionPolicy` is even needed anymore or if there are other objects that I have missed which would replace the functionality of `SmartRetryExecutionPolicy` object?

nozzlegear:
Hey! The LeakyBucketExecutionPolicy is our replacement for the SmartRetryExecutionPolicy. It should handle everything the old policy did, along with handling graphQL requests and other edge cases.

--
Josh Harms

PR #923, opened 2023-08-24
Update retry policy name in README to reflect current version of library

Terit:
When I attempted to implement a global retry policy I couldn't figure out why the README's code wasn't working so I went looking for the `SmartRetryExecutionPolicy` in the codebase only to find issue #853. 

I've updated the README to reference the replacement policy. I didn't dig into `LeakyBucketExecutionPolicy` to see if the second statement about infinite retries still applies or not.

> Keep in mind that the `RetryExecutionPolicy` and the `LeakyBucketExecutionPolicy` will keep retrying your requests – potentially until the end of time – until they are successful. It's up to you to ensure that such a strategy won't impact the performance of your applications.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -2874,7 +2874,7 @@ However, ShopifySharp also has request execution policies that you can use to im
 
 1. `DefaultRequestExecutionPolicy`: This is the default policy, which will throw a `ShopifyRateLimitException` when the API rate limit has been reached.
 2. `RetryExecutionPolicy`: If a request throws a `ShopifyRateLimitException`, this policy will keep retrying it until it is successful.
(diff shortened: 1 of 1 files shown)


nozzlegear:
Good catch, thanks! The docs are desperately in need of an update. I keep telling myself I want to put together a nice website and Github wiki for them, but I have never managed to find the time to do that.

As far as the infinite retries thing, I believe that's still correct for the Leaky Bucket policy.

clement911:
It's still correct indeed, but it only retries for rate limit errors. If any other type of error is encountered, the error bubbles up immediately as an exception.

PR #934, opened 2023-10-03
Re-added removed method useful to deserialize webhook payload into ob…

clement911:
…jects.


nozzlegear:
Thanks! Did I remove that one? I think I did but can't remember why I would've done that.

clement911:
@nozzlegear it was me that had removed this one.
It's a method not used by ShopifySharp but that can be used by users to deserialize webhooks.
it was not released yet so nothing would have been broken.
I triggered another release with this change, which is pending your review.

Issue #933, opened 2023-09-29
SmartCollectionRules: Add ConditionObjectId Property


alonghurst:
Hi

I'm trying to create a `SmartCollection` via the `SmartCollectionService` but I'm having trouble setting `SmartCollection.Rules` to use a value from a product metafield.

When I set the properties on newly created `SmartCollectionRules` objects the API returns `'(422 Unprocessable Entity) base: metafield definition id is invalid or missing.`.

The Shopify documentation (https://shopify.dev/docs/api/admin-rest/2023-07/resources/smartcollection) refers to a `condition_object_id` property:

_condition_object_id: The object id that points to additional attributes for the collection rule. This is only required when using metafield definition rules._

Adding the following property to `SmartCollectionRules` has solved the issue for me (both read and write of SmartRules referring to product metafields):

```
[JsonProperty("condition_object_id")] 
public long? ConditionObjectId { get; set; }
```

Do you have any plans to add a feature like this into an upcoming release? Would you like me to raise a PR for the change?

Thanks

alonghurst:
@nozzlegear want me to raise a PR for this change?

nozzlegear:
@alonghurst Sure! That'd be great.

PR #949, opened 2023-10-26
Added SmartCollectionRules.ConditionObjectId

alonghurst:
https://github.com/nozzlegear/ShopifySharp/issues/933

This change allows reading/writing of `SmartCollectionRules` which use metafields in `SmartCollections`.

Shopify Docs: https://shopify.dev/docs/api/admin-rest/2023-07/resources/smartcollection

It can be used to create a metafield rule in a smart collection as follows (example shows for a product metafield):

```
var rule = new SmartCollectionRule
{
    Column = "product_metafield_definition",
    Relation = "equals",
    Condition = value
    ConditionObjectId = conditionObjectId
};
```

I found the most straightforward way to find the `ConditionObjectId` for a given product metafield was to inspect the HTML of the Settings / Custom Data page in my store and find the ids there:

`<li id="gid://shopify/MetafieldDefinition/123456789" class="Polaris-IndexTable__TableRow_1a85o">`

where `123456789` is to be used as `conditionObjectId`.

### DIFF
--- a/ShopifySharp/Entities/SmartCollectionRules.cs
+++ b/ShopifySharp/Entities/SmartCollectionRules.cs
@@ -24,5 +24,11 @@ namespace ShopifySharp
         /// </summary>
         [JsonProperty("column")]
         public string Column { get; set; }
+
+        /// <summary>
+        /// The id of the object that points to additional attributes for the collection rule. This is only required when using metafield definition rules.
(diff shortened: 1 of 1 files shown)


nozzlegear:
Published in 6.6.1 on Nuget. Thanks for contributing!

PR #963, opened 2023-12-04
Add ShopifySharp.Extensions.DependencyInjection project

nozzlegear:
I'm working on adding a ShopifySharp.Extensions.DependencyInjection package in this pull request. The idea is that, for every service interface in ShopifySharp there will be a matching service "factory" interface that can be dependency injected like this:

```cs
public class OrdersController(IOrderServiceFactory orderServiceFactory) : Controller
{
    [HttpGet("order/{id}")]
    public async Task<ActionResult> GetOrder([FromRoute] long id)
    {
        var user = await DoSomethingToGetUser();
        var credentials = new ShopifySharp.ShopifyRestApiCredentials(user.ShopDomain, user.AccessToken);
        var orderService = orderServiceFactory.Create(credentials);
        var order = await orderService.GetAsync(id);

        return View(order);
    }
}
``` 

Adding the factories to the DI container can be done like this:

```cs
public void ConfigureServices(IServiceCollection services)
{
    services.AddShopifySharp();
}
```

Which will add every factory as a singleton to your DI container. The factories are configured to take a `IRequestExecutionPolicy` from the DI container as well, if you set one up:

```cs
public void ConfigureServices(IServiceCollection services)
{
    services.AddSingleton<IRequestExecutionPolicy, LeakyBucketExecutionPolicy>();
    services.AddShopifySharp();
}
```

If you only want to add certain factories to the container, you can skip the `AddShopifySharp` call and just add the factories you need:

```cs
public void ConfigureServices(IServiceCollection services)
{
    services.AddSingleton<IRequestExecutionPolicy, LeakyBucketExecutionPolicy>();
    services.AddSingleton<IOrderServiceFactory, OrderServiceFactory>();
    services.AddSingleton<ICustomerServiceFactory, CustomerServiceFactory>();
}
```

I'm open to feedback on the implementation! In a future pull request I want to add support for custom HttpClients.

One thing that I do plan on doing before merging this is moving all of the factories to the ShopifySharp package itself, and just keeping the DI setup/extension methods in the ShopifySharp.Extensions.DependencyInjection package. I think the factories are useful enough to have in ShopifySharp proper.

### TODO

- [x] Move the factories into ShopifySharp proper.
- [x] Add Nuget package publishing to the Github Actions deployment process. 
    - [x] Only run this when something in the DI project has changed. [1] 
    
[1] A future improvement would run the `generate_factories.fish` script on the ShopifySharp project's builds and compare its output to previous runs, then publish the DI project if factories have changed.


ShopifySharp/ShopifySharp.csproj
```diff
@@ -1,15 +1,13 @@
...
 		<AssemblyName>ShopifySharp</AssemblyName>
 		<RootNamespace>ShopifySharp</RootNamespace>
 		<VersionPrefix>6.6.1</VersionPrefix>
 		<PublishRepositoryUrl>true</PublishRepositoryUrl>
 		<IncludeSymbols>true</IncludeSymbols>
 		<SymbolPackageFormat>snupkg</SymbolPackageFormat>
-	</PropertyGroup>
-	<PropertyGroup Condition="'$(TargetFramework)' == 'net6.0'">
-		<LangVersion>9.0</LangVersion>
+        <LangVersion>Latest</LangVersion>
```
nozzlegear, line 10: I didn't mean to merge this in, but I actually want to confirm this: I'm able to build the netstandard2.0 project using nullable references and `<LangVersion>Latest</LangVersion>`. Will these nullable references work in .NET Framework? Even if they don't, if the package itself still continues to function the same in DNF but gives everything else nullable references, that would be a huge improvement.

nozzlegear: I'm going to risk it and merge this in, since I have no way to test DNF on a Mac. We'll let the test suite run and see what happens.

PR #965, opened 2023-12-06
Adds two overloads to GraphService.SendAsync to allow passing mediaType for application/json formatted queries.

jakenuts:
> I'm absurdly new to GraphQL and may not fully understand what it is at play in passing the application/graphql vs application/json media types but as I was hoping to create a parameterized query and it worked with Json.Net I was hoping to make the same possible for the System.Text.Json variants. Please feel free to burn this, salt the ground where it was and move on if it serves no purpose.

1. Added overloads of the `GraphService.SendAsync` methods to allow passing `mediaType` parameter so "application/json" can be used when passing a json query and variables.

2. Forwarding existing  `GraphService.SendAsync` to the overloads passing the "application/graphql" media type to preserve their behavior.

3. Added test method `GetProductsAsParameterizedJsonQuery`  to validate the new SendAsync overloads.

### DIFF
--- a/ShopifySharp.Tests/GenerateGraphQLSchema_Test.cs
+++ b/ShopifySharp.Tests/GenerateGraphQLSchema_Test.cs
@@ -92,6 +92,35 @@ namespace ShopifySharp.Tests
                 Assert.Null(commentEventEmbed.AsCustomer());
             }
         }
+
+        [Fact]
+        public async Task GetProductsAsParameterizedJsonQuery()
+        {
+            var svc = new GraphService(Utils.MyShopifyUrl, Utils.AccessToken);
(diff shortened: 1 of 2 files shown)


clement911:
Hi @jakenuts I see how this can be useful to allow passing variables.

There was a few things I wanted to change and clean up so I've made another PR and will close this one.

See https://github.com/nozzlegear/ShopifySharp/pull/966

In particular, the new PR makes it easier to use because the user doesn't have to serialize the request themselves, or pass a media type.

Let me know if you have any remarks though.

jakenuts:
Great, I was hesitant about the asymmetry of the methods, glad you found a way to make it more consistent. Thanks!

PR #967, opened 2023-12-08
Added missing 'Currency' property to 'Charge' entity. 

ltasler:
I've added `currency` property to `Charge` entity. This is taken from: https://shopify.dev/docs/api/admin-rest/2023-10/resources/applicationcharge


nozzlegear:
Awesome, thanks! I'm publishing a release today, I'll include this.

PR #977, opened 2023-12-15
Rename ShopifyRestApiCredentials to ShopifyApiCredentials

nozzlegear:
This pull request renames the `ShopifyRestApiCredentials` struct to `ShopifyApiCredentials`. I thought I was going to need three to four different structs to hold different credential types, but it turns out we really only need two: one for the partner credentials (organization id + access token), and one for basically everything else (shop domain + access token).

That in mind, I'm going to quickly rename the struct to `ShopifyApiCredentials` while it's brand new so it can be used with the graph service without feeling awkward.

### DIFF
--- a/ShopifySharp/Credentials/ShopifyApiCredentials.cs
+++ b/ShopifySharp/Credentials/ShopifyApiCredentials.cs
@@ -7,7 +7,7 @@ using System;
 
 namespace ShopifySharp.Credentials
 {
-    public readonly struct ShopifyRestApiCredentials(string shopDomain, string accessToken)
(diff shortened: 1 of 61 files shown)


ShopifySharp/Credentials/ShopifyApiCredentials.cs
```diff
@@ -36,5 +36,15 @@ public override int GetHashCode()
             return HashCode.Combine(ShopDomain, AccessToken);
             #endif
         }
+
+        public static bool operator ==(ShopifyApiCredentials left, ShopifyApiCredentials right)
```
nozzlegear, line 40: I also added some extra `credentials == otherCredentials` equality operators.

Issue #962, opened 2023-12-03
Listing products in English 


jakenuts:
I tried adding /en to my shopify url but it gets stripped out in BuildShopUri. Is there any way to pass the language path to ShopifySharp such that calls like ListAsync use the /en path in the request?

nozzlegear:
Hey! That behavior is expected based on how I wrote the base ShopifyService class. I might be wrong, but I don't think the REST API can be localized using those language paths like the storefront API can. Happy to be wrong about that though, I'll give it a test and see what happens.

jakenuts:
Oh, that would make sense. Thanks!

jakenuts:
Sorry to bother you again, but I was wondering how you generated the graphql entities? When I learned that I would need to switch apis to get translations I went looking for a graphql schema for products/variants/related to feed into various tools and all my attempts at introspection produce massive files with some special (non-standard?) bits that break each tool in their own way (ZeroQL, Apollo, etc). 

Thanks!

nozzlegear:
Ah actually all of the generated schemas are handled by Clement. @clement911 would you mind giving a brief overview on how you generate those schemas? I think you explained it somewhere here in the repo but I haven't been able to find it back, and I've been meaning to ask you just for my own curiousity.

nozzlegear:
Ah that’s awesome, thanks! I could definitely use that in other projects. I’ll set up a Github Action so you and I can trigger the regeneration from the actions page, so you don’t have to run the test manually to get the changes.

--
Joshua Harms

clement911:
Oh ok, I'm not too sure how you will set this action but honestly using the unit test to trigger the regeneration has been working quite well. All you need to do is remove the `Skip` attribute and run it. Then add the `Skip` attribute again so that it doesn't run on each CI build.

nozzlegear:
I think I can do it if I mess around with the test category, like we use to exclude/include dotnet framework tests. I'll give it a shot tonight and create a pull request for it.

@jakenuts Let us know if you have any questions about the schema generation!

jakenuts:
That absolutely does, thank you @clement911  & @nozzlegear 😊

PR #1011, opened 2024-02-05
Release 2024-01

nozzlegear:
Releasing 7.0 is going to take a little longer than I thought, so I’m going to push out 2024-01 as part of 6.13.0.

I realize this contains breaking changes for the  generated GraphQL types, but I think I’m okay with that. Since ShopifySharp doesn’t version itself the same way Shopify versions their API, it’s bound to happen and I don’t want to release a new major version each time they deprecate or change their graph (or rest) types.


clement911:
I personally think it's ok to increment the major version regularly.
As an example, I use [Stripe](https://github.com/stripe/stripe-dotnet) quite a bit, and they are already at major version 43!

Here we could increment to 7. And when you do your other changes, we could increment to 8, even if it's shortly after...

nozzlegear:
Oh wow, 43 is crazy! Each major increment is breaking for them too?

--
Joshua Harms

clement911:
Yes I believe they follow semver like most libraries. So each major increment is a breaking change.
But it doesn't mean that it's hard to upgrade. Just that there is at least one breaking change.

adearriba:
QQ: How are you guys generating the graphql types?

clement911:
At a high level, the code generation of graphql type uses a graphql introspection query to load all types and generate some c# code.
It can be triggered by running the test in (after removing the Skip attribute) https://github.com/nozzlegear/ShopifySharp/blob/master/ShopifySharp.Tests/GenerateGraphQLSchema_Test.cs

adearriba:
Thanks @clement911 I was asking more on which code generator you guys were using for this.

nozzlegear:
We're using a generator that Clement wrote called [Wish.GraphQLSchemaGenerator](https://github.com/Wish-Org/Wish.GraphQLSchemaGenerator). He explained it here in #962!

Issue #1009, opened 2024-02-03
Regarding the Dispute of Shopify Payment


smart222zx:
Hello, I couldn't find any services related to Dispatch Evidence and Dispatch File Upload, but the documentation includes these APIs.

Docs:
[https://shopify.dev/docs/api/admin-rest/2024-01/resources/dispute-evidence](url)
[https://shopify.dev/docs/api/admin-rest/2024-01/resources/dispute-evidence](url)

nozzlegear:
Thanks @smart222zx! If you're interested in writing `DisputeEvidenceService` and opening a pull request for it here on Github, I'd be happy to review it and merge it into ShopifySharp. Otherwise I'll get these APIs added as soon as I have some extra time!

PR #1025, opened 2024-02-22
Feature/dispute evidence

dsargent9182:
https://github.com/nozzlegear/ShopifySharp/issues/1009
I have attempted to implement this feature. I was able to pull a dispute but when I tried to pull dispute evidence it would always return a 404. I can see from the documentation it says I need the scope -> shopify_payments_dispute_evidences but I don't see how to enable this. 
DisputeEvidence

### DIFF
--- a/ShopifySharp/Entities/ShopifyPaymentsDisputeEvidence.cs
+++ b/ShopifySharp/Entities/ShopifyPaymentsDisputeEvidence.cs
@@ -7,63 +7,63 @@ using System.Threading.Tasks;
 
 namespace ShopifySharp
 {
-	/// <summary>
-	/// An object representing a Shopify payment dispute evidence.
-	/// </summary>
-	public class ShopifyPaymentsDisputeEvidence : ShopifyObject
-	{
-		[JsonProperty("payments_dispute_id")]
-		public long? PaymentsDisputeId { get; set; }
+    /// <summary>
+    /// An object representing a Shopify payment dispute evidence.
+    /// </summary>
+    public class ShopifyPaymentsDisputeEvidence : ShopifyObject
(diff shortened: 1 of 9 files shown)


ShopifySharp/Entities/ShopifyPaymentsDisputeEvidence.cs
```diff
@@ -0,0 +1,69 @@
...
+        public DateTimeOffset? UpdatedOn { get; set; }
+
+        [JsonProperty("dispute_evidence_files")]
+        public ShopifyPaymentsDisputeEvidenceFiles DisputeEvidenceFiles { get; set; }
+
+        [JsonProperty("billing_address")]
+        public ShopifyPaymentsDisputeEvidenceAddress BillingAddress { get; set; }
+
+        [JsonProperty("fulfillments")]
+        public ShopifyPaymentsDisputeEvidenceFulfillment[] Fulfillments { get; set; }
```
nozzlegear, line 67: This should be an `IEnumerable<ShopifyPaymentsDisputeEvidenceFulfillment>` to match the rest of the library.

nozzlegear (changes requested):
Thanks for the pull request @dsargent9182! I've requested just one small change to make one of the new classes follow the style used by the rest of the package. Other than that, it all looks good to me!

ShopifySharp/Entities/ShopifyPaymentsDisputeEvidence.cs
dsargent9182: @nozzlegear Thanks I have made that update.

nozzlegear (approved):

PR #1034, opened 2024-02-25
Skipping failing test that goes into infinite loop

clement911:
@nozzlegear  I had to skip this test because it doesn't work anymore.
This caused the build to go into an infinite loop.


nozzlegear:
Interesting , thanks! I wonder if they changed something recently. I’ve been planning to update tests like this one to use mocks and give it the json we expect Shopify to return, maybe I’ll do that for this case to ensure the test is still testing the error parsing.

Glad you found the issue so quickly!

clement911:
Also I noticed that this test was calling the API without any API version in the URL, so Shopify defaults to the oldest supported version.

PR #1043, opened 2024-03-21
Update readme.md

VincentMatthews:
Extended ShopifySharp versions to include Shopify API version 2024-01


nozzlegear:
Good catch, thanks a bunch @VincentMatthews!

PR #1051, opened 2024-04-05
Deprecate most GraphService methods and converge on two standard methods

nozzlegear:
This change deprecates most GraphService.PostAsync and GraphService.SendAsync methods, of which there were at least six total. It introduces two new GraphService.PostAsync methods to replace them all, which standardize the usage of the `GraphRequest` class and serializing all requests to json.

Previously, some of the PostAsync/SendAsync methods would serialize the request to json, some wouldn't. Some would send the variables, some wouldn't. Some would let you set your own json wherein you could apply the variables yourself, but then they'd send the `application/graphql` content-type, which made Shopify ignore any variables sent. 

That one has personally bit me each time I've tried to use the GraphService recently; it's been difficult to figure out *which* method should be used on the service when I want to send variables, without looking at the source code for ShopifySharp.

After this change, the new method for sending Graph requests using the GraphService going forward would look like this:

```csharp
var query = 
"""
query GetStuff($productId: String!) {
  product(id: $productId) {
    id
    title
  }
}
""";
var variables = new
{
  productId = "gid://shopify/products/123456"
};
var graphRequest = new GraphRequest
{
  Query = query,
  Variables = variables,
  EstimatedGraphCost = 3
};

// Get a JsonElement back
var result = await graphService.PostAsync(graphRequest, cancellationToken);
// Or deserialize into whatever you want
var result = await graphResult.PostAsync<ShopifySharp.GraphQL.Product>(graphRequest, cancellationToken);
```

I don't want anyone to lose the ability to control how variables are serialized or deserialized though, so I've added an `IGraphSerializer` interface that can be passed to the GraphService constructor. 

```csharp
public class MyCustomGraphSerializer : IGraphSerializer
{
  public string SerializeToJson(IDictionary<string, object> variables)
  {
    throw new NotImplementedException();
  }
  
  public T DeserializeFromJson<T>(string json)
  {
    throw new NotImplementedException();
  }
}

// elsewhere
var serializer = new MyCustomGraphSerializer()
var graphService = new GraphService(domain, accessToken, graphSerializer: serializer);
```

I'm not 100% sure this is the exact interface I'll use for this, but it's something along the lines of what I'm thinking. Whether we use this interface or do something else for serialization/deserialization in the GraphService, I want it to play nicely with the factory and DI package.


clement911:
Should it be also be possible to provide variables via a dictionary?

nozzlegear:
Oh yeah, I like that more @clement911! So maybe we change the GraphRequest class to look like this:


```cs
public class GraphRequest
{
  public string Query { get; set; }
  public Dictionary<string, object> Variables { get; set; }
  
  // Convenience methods for adding/setting variables?
  public GraphRequest AddVariable(string key, object value)
  {
    Variables.Add(key, value);
    return this;
  }
  
  public GraphRequest SetVariable(string key, object value)
  {
    Variables.Set(key, value);
    return this;
  }
}
```

Then you could just pass in an entire dictionary if you've already got one, use the convenience methods, or access the dictionary object directly.

clement911:
I'm not sure that the `AddVariable` / `SetVariable` methods really are needed.
The user can simply use the `Variables` dictionary to manage variables, right?

nozzlegear:
Agreed, I was thinking about that too and it doesn’t sound that useful. I’m going to drop that idea, but I did make the change to switch the Variables type to a dictionary.—Joshua HarmsOn Apr 7, 2024, at 17:19, Clement Gutel ***@***.***> wrote:﻿
I'm not sure that the AddVariable / SetVariable methods really are needed.
The user can simply use the Variables dictionary to manage variables, right?

nozzlegear:
Remaining tasks:

- [ ] Unit tests to ensure that obsoleted methods did not break with this PR.
- [ ] Unit tests to ensure proper serialization of the GraphRequest, GraphExtensions and GraphUserError objects.
- [ ] Unit tests for disabling error handling in the GraphService
- [ ] Documentation for this release

clement911:
@nozzlegear I know that you are actively working on this and wondered if you saw this: https://shopify.dev/changelog/content-type-application-graphql-is-deprecated

@matnun-br FYI

nozzlegear:
@clement911 Oh yes, I did see that! Thanks for the reminder, I want to incorporate that.

nozzlegear:
Okay, I've got almost all of the changes in that I wanted to for this PR. It definitely crept in scope but I'm pretty happy with it. The GraphService class is sitting at 100% test coverage right now and I'm aiming to maintain that going forward.

Just two things I want to squeeze in before merging:

- [ ] Merge `PostAsync(GraphRequest graphRequest, Type returnType)` from the master branch.
- [ ] Add and support enum `GraphExtensionsHandling` (name tbd) that tells the service whether it should look for `"extensions"` property on the returned graph response and try to parse it into `GraphResponse.Extensions`.

nozzlegear:
Merged! 🎉 I'll check the tests, draft a new Github/Nuget release and begin updating the documentation tomorrow afternoon.

dg107:
I'm sure I'm the last visual basic programmer on the planet, but now this class is crashing as VB can't handle there being a query property and a Query property.

nozzlegear:
@dg107 Oh, I didn't know that. Thanks for the heads up! I mostly use F# in my personal projects so I know the pain of devs assuming if something works in C# it must work in everything else. 😜 

I'll do some digging and see if I can't get this fixed somehow.

nozzlegear:
Hey @dg107 (and @CopilotDaniel) – I've created a separate issue for the Visual Basic problem. I'm not sure if you've seen it, can you check the solution I've proposed there and let me know if it'll work for you?

Here's the link: https://github.com/nozzlegear/ShopifySharp/issues/1144

PR #1092, opened 2024-08-20
Feature/wiki in repo

ajvanlaningham:
creates the wiki docs reference as a submodule of the main library under docs/wiki
Also updates wiki metafield.md file
Commits accidentally made with my work account


nozzlegear:
I thought I had merged this right away. Sorry about that – thank you!

PR #1113, opened 2024-11-29
Skipping location tests because read_locations scope is required as of 2024-10

clement911:
Skipping location tests because read_locations scope is required as of 2024-10

### DIFF
--- a/ShopifySharp.Tests/Location_Tests.cs
+++ b/ShopifySharp.Tests/Location_Tests.cs
@@ -14,7 +14,7 @@ public class Location_Tests
         Service.SetExecutionPolicy(new LeakyBucketExecutionPolicy());
     }
 
-    [Fact]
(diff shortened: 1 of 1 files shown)


clement911:
FYI @nozzlegear , can you change the token used to run the tests to ensure the read_locations was granted?
Starting 2024-10, the Location REST API is gated by read_locations

nozzlegear:
Oh right, I thought I had done that already but may have only been for the one on my local machine. I'll add it to my todo list and get it done tomorrow.-- Joshua HarmsOn Nov 28, 2024, at 19:37, Clement Gutel ***@***.***> wrote:﻿
FYI @nozzlegear , can you change the token used to run the tests to ensure the read_locations was granted?
Starting 2024-10, the Location REST API is gated by read_locations

PR #1117, opened 2024-12-10
Add non generic version of GraphService.SendAsync()

clement911:


clement911:
Thanks for the heads up

PR #1199, opened 2025-07-02
Add CurrentQuantity, Gets or sets the current quantity of the item (I…

joepperkins:
…f the Quantity is ajusted after the order is created this will show the current quantity).


nozzlegear:
Sorry @joepperkins, I missed your pull request! I need to set up a bot to ping me when things sit for too long without my attention. I'll get this published today!

PR #1215, opened 2025-09-10
Update Collect.Position to a long

nate141:
We started getting multiple errors from different integrations: 
```
Could not convert to integer: 158963409999360. Path 'collects[0].position' 
Could not convert to integer: 2694011473488384. Path 'collects[0].position'.
```
And more

Seems like Shopify started returning a long for this field too


nate141:
Thanks for merging!

Any clue when I can expect an updated NuGet package?

Thanks again

PR #1259, opened 2026-07-15
Add support for querying interfaces and operations which return interfaces

nozzlegear:
This PR adds the necessary parsing and infrastructure to make it possible to generate new `InterfaceCasesBuilderBase` query builders on any interface type with concrete implementations, or any operation that returns an interface with concrete implementations. Now, ShopifySharp users will be able to write queries in .NET for Shopify interfaces which will deserialize to the correct concrete type (using the same `... on ConcreteTypeName { }` semantics as the union cases builders).

Of particular note, this opens up the `nodes(){ }` and `node() { }` queries (`NodesOperationQueryBuilder` and `NodeOperationQueryBuilder`, respectively), which were previously just empty classes that had no methods on them as ShopifySharp was unable to handle the `INode` return type. This change will now let us use `query.Node(x => x.OnSomeType(t => t.Foo()))`.


ShopifySharp/GraphQL/QueryBuilders/InterfaceCasesBuilderBase.cs
```diff
@@ -0,0 +1,19 @@
...
+
+[PublicAPI]
+public abstract class InterfaceCasesBuilderBase<TSource, TSelf>(IQuery<TSource> query): IUnionCasesBuilder<TSelf>
+    where TSelf : IUnionCasesBuilder<TSelf>
+{
+    protected readonly IQuery<TSource> InnerQuery = query;
+
+    protected abstract TSelf Self { get; }
+
+    public TSelf AddUnionCase<TUnionCase>(IQuery<TUnionCase> unionCaseQuery) where TUnionCase : class?
```
nozzlegear: I need to rename this method – these aren't union cases, they're interfaces.

ShopifySharp.GraphQL.Parser/UnionsBuilderWriter.fs
```diff
@@ -36,15 +36,17 @@ type UnionsBuilderWriter(type': VisitedTypes, context: IParsedContext) =
...
                 collectUnionFields visitedType
-            | ReturnType.FieldType (FieldType.ValueType (FieldValueType.GraphObjectType fieldValueType) as fieldType) ->
-                match AstNodeMapper.unwrapFieldType fieldType with
+            | ReturnType.FieldType fieldType ->
+                let unwrapped = AstNodeMapper.unwrapFieldType fieldType
+                match unwrapped with
                 | FieldValueType.GraphObjectType (NamedType.UnionType (_, unionCaseNames)) ->
-                    // TODO: naming is probably janky here
-                    [| UnionCasesBuilderWriter(type', fieldValueType.Name, unionCaseNames, None, context)|]
+                    let name =
```
nozzlegear: Not sure what I was doing here, but there's no need to match the thing we're already matching just to get its name.

PR #1261, opened 2026-07-21
Fix refresh request

GMatrixGames:
Not sure how this got past me, but I'm guessing I misread something because the original request body for refreshing was not right.


nozzlegear:
Nice catch! I'm working on ShopifySharp right now, planning to release it this evening!

PR #1262, opened 2026-07-24
Add ShopifyOauthUtility method overloads, deprecate old overloads, use custom exception

nozzlegear:
Various changes refactoring the ShopifyOauthUtility class:

- Added a new Type field to the AuthorizationResult, which indicates whether the token is a legacy permanent offline token; an expiring offline token (new); or a store user token. Defaults to the legacy offline token if not explicitly set.
- Renamed the RefreshAccessTokenIfNeededAsync method to RefreshExpiringTokenIfStaleAsync
- The RefreshExpiringTokenIfStaleAsync method will check the AuthorizationResult's token type and throw if it's not an expiring token.
- Added an overload for that method which accepts a pared down options type where callers only need to pass the refresh token string and some optional datetime values instead of an AuthorizationResult.
- Deprecated a couple of existing RefreshAccessTokenAsync methods; one because it had 4-5 string parameters (getting too unwieldy), and one because the "RefreshExpiring" name is more clear about which kind of token it accepts.
- Added an internal ArgumentException.ThrowIfNull extension method for netstandard2.0.


ShopifySharp.Tests/Utilities/ShopifyOauthUtilityTests.cs
```diff
@@ -44,6 +44,8 @@ public ShopifyOauthUtilityTests()
             .Returns(httpClientFactory);
         A.CallTo(() => _serviceProvider.GetService(typeof(IJsonSerializer)))
             .Returns(_jsonSerializer);
+        A.CallTo(() => _serviceProvider.GetService(typeof(TimeProvider)))
+            .Returns(TimeProvider.System);
```
nozzlegear: This needs to be a FakeTimeProvider when used in unit tests.

ShopifySharp.Tests/Utilities/ShopifyOauthUtilityTests.cs
```diff
@@ -1267,6 +1280,271 @@ public async Task AuthorizeAsync_WhenAssociatedUserScopeIsAnInvalidType_ShouldTh
...
+        };
+        expiringOfflineToken.Type.Should().Be(ShopifyAccessTokenType.ExpiringOffline);
+
+        // Legacy permanent offline access token
+        var permanentOfflineToken = new AuthorizationResult("token", []);
+        permanentOfflineToken.Type.Should().Be(ShopifyAccessTokenType.LegacyPermanentOffline);
+    }
+
+    [Fact]
+    public void AuthorizationResult_ExpiresIn_ShouldFallbackToOnlineAccessExpiresIn()
```
nozzlegear, line 1306: This test, and those that follow it, should use the FakeTimeProvider to create timespans/datetimes.

nozzlegear (commented):
Need to make a couple changes to the tests before I merge this.

PR #1268, opened 2026-09-15
Add support for the client credentials grant OAuth flow

### LINKED ISSUE #1225: Request access token for custom apps created using Dev Dashboard from 2026
Can you please add a option to request a access token for custom apps created using the new Dev Dashboard?
https://shopify.dev/docs/apps/build/authentication-authorization/access-tokens/client-credentials-grant

I have tried using the `ShopifyOauthUtility.AuthorizeAsync`, but it requires a code.

https://help.shopify.com/en/manual/apps/app-types/custom-apps
> Starting from January 1, 2026, you can't create any more legacy custom apps. To create and manage new apps, you need to use the [Dev Dashboard](https://shopify.dev/docs/apps/build/dev-dashboard). This change doesn't impact any of your existing custom apps.

See related [community post here](https://community.shopify.dev/t/how-to-use-shopify-s-api-with-custom-app-changes/24852/2)

nozzlegear:
This commit adds a new `GetClientCredentialsAccessTokenAsync` method to the `ShopifyOauthUtility`, which implements the client credentials grant flow required by custom Shopify apps. These client credentials expire after 24 hours, but the flow does not return a refresh token – instead, devs just call the same grant method again, which automatically refreshes the token.

Closes #1225


ShopifySharp/Utilities/ClientCredentialsAccessTokenOptions.cs
```diff
@@ -0,0 +1,26 @@
...
+public record ClientCredentialsAccessTokenOptions
+{
+    /// The store's *.myshopify.com URL.
+    public
+#if NET6_0_OR_GREATER
+    required
+#endif
+    string ShopDomain { get; set; } = null!;
+
+    /// Your app's public Client ID, also known as its public API key.
```
nozzlegear, line 13: "public API key" is probably confusing when we're talking about custom app, I should reword this.

ShopifySharp/Utilities/ShopifyOauthUtility.cs
```diff
@@ -408,15 +438,13 @@ public async Task<AuthorizationResult> RefreshOfflineAccessTokenIfStaleAsync(
...
         ArgumentNullException.ThrowIfNull(options);
 
+
         if (currentResult.RefreshTokenHasExpired())
             throw new ShopifyInvalidRefreshTokenException("The authorization result's refresh token has expired and can no longer be used to refresh the access token.");
 
         if (!currentResult.ShouldRefreshAccessToken(options.RefreshBeforeExpiry))
             return currentResult;
 
-        if (!currentResult.HasRefreshToken)
```
nozzlegear, line 417: I think I already checked that removing this was fine, but should double-check. Is the AuthorizationResult ensuring we guard against this scenario already?

